PR #76, opened 2019-02-07
Add link to SwiftFormat’s redundantSelf rule

nicklockwood:
#### Summary

Added link to SwiftFormat rule for removing implicit self.

#### Reasoning

SwiftFormat is able to reliably remove all redundant uses of `self` automatically (SwiftLint can automatically insert `self`, but not remove it).

#### Reviewers

### DIFF
--- a/resources/airbnb.swiftformat
+++ b/resources/airbnb.swiftformat
@@ -1,7 +1,8 @@
 # options
+--self remove # redundantSelf
 --importgrouping testable-bottom # sortedImports
 --commas always # trailingCommas
 --trimwhitespace always # trailingSpace
 
 # rules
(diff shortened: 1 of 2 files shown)


fdiaz:
@nicklockwood could you also add the rule to this file? https://github.com/airbnb/swift/blob/master/resources/airbnb.swiftformat

resources/airbnb.swiftformat
```diff
@@ -1,7 +1,8 @@
 # options
 --importgrouping testable-bottom # sortedImports
 --commas always # trailingCommas
+--self remove # redundantSelf
```
fdiaz: Could we move this to the top so this is sorted alphabetically by rule?

PR #78, opened 2019-02-08
Remove redundant guideline

nicklockwood:
#### Summary

Removed the guideline about `where`-clause positioning in `switch` statements.

#### Reasoning

Swift already generates a warning for this scenario (see screenshot), so mentioning it seems redundant (on the assumption that "code should compile with no warnings" is already an implicit guideline).

screen shot 2019-02-07 at 23 53 46

#### Reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -1044,35 +1044,6 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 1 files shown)


dcramps:
👍 to this. When I added this rule, the compiler warning was not there.

Issue #25, opened 2018-10-31
Indentation Rule Conflicts with Xcode Indentation Behavior


brynbodayle:
For this rule:

> Closure end braces should have the same indentation as the line with their opening brace. This makes it easier to follow control flow through closures.

https://github.com/airbnb/swift#closure-end-brace-indentation

It's in conflict with this guiding tenet:

> These rules should not fight Xcode's ^ + I indentation behavior.

Style guide example:

```
// RIGHT

match(pattern: pattern)
  .compactMap { range in
    return Command(string: contents, range: range)
  }
  .compactMap { command in
    return command.expand()
  }
```

Xcode indentation:

```
// RIGHT

match(pattern: pattern)
  .compactMap { range in
    return Command(string: contents, range: range)
  }
  .compactMap { command in
    return command.expand()
}
```

calda:
Closing this one since it seems outdated -- Xcode indentation is quite a moving target 😅

PR #80, opened 2019-02-08
Remove closure_end_indentation rule

rholmes:
#### Summary

Remove the closure_end_indentation rule

#### Reasoning

It violates the second guiding tenet of the style guide:
> These rules should not fight Xcode's ^ + I indentation behavior.

image

I understand that many people dislike Xcode's default style of indenting closure end braces, but I think the guiding tenet is more important. This exception will negatively impact programmer experience and productivity by guaranteeing a locally inconsistent style and linting failure every time Xcode autocompletes a block or we use ^+I.  

The "don't fight Xcode" tenet is important to me because it, in turn, reflects a much broader tenet: "do no harm". It's an assurance that the goal of the style guide is to assist and supplement our other tools to improve the clarity and consistency of code -- not to impose rules even if there are negative side effects.

P.S. `SwiftLint` is new to me so I don't know if there's a way to modify the rule to reflect Xcode's default indentation and leave the door open to change the details later (e.g. if Xcode ever makes this configurable).

#### Reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,42 +593,6 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
-* <a id='closure-end-brace-indentation'></a>(<a href='#closure-end-brace-indentation'>link</a>) **Closure end braces should have the same indentation as the line with their opening brace.** This makes it easier to follow control flow through closures. [![SwiftLint: closure_end_indentation](https://img.shields.io/badge/SwiftLint-closure__end__indentation-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#closure-end-indentation)
-
-  <details>
-
-  ```swift
-  // WRONG
-
-  match(pattern: pattern)
-    .compactMap { range in
-      return Command(string: contents, range: range)
-  }
-    .compactMap { command in
-      return command.expand()
-  }
-
-  values.forEach { value in
-    print(value)
-    }
-
-  // RIGHT
-
-  match(pattern: pattern)
-    .compactMap { range in
-      return Command(string: contents, range: range)
-    }
-    .compactMap { command in
-      return command.expand()
-    }
-
-  values.forEach { value in
-    print(value)
-  }
-  ```
-
-  </details>
-
 * <a id='closure-brace-spacing'></a>(<a href='#closure-brace-spacing'>link</a>) **Single-line closures should have a space inside each brace.** [![SwiftLint: closure_spacing](https://img.shields.io/badge/SwiftLint-closure__spacing-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#closure-spacing)
 
   <details>
--- a/resources/swiftlint.yml
+++ b/resources/swiftlint.yml
@@ -1,5 +1,4 @@
 whitelist_rules:
-  - closure_end_indentation
   - closure_spacing
   - colon
   - control_statement


bachand:
I prefer the indentation of this rule to what happens in Xcode by default. However I would defer to the guiding tenet here as well.

erichoracek:
@rholmes can we update the indentation of our array examples to reflect these changes as well? e.g.: 
https://github.com/airbnb/swift/blob/master/README.md#trailing-comma-array
```diff
 // RIGHT
 let rowContent = [
   listingUrgencyDatesRowContent(),
   listingUrgencyBookedRowContent(),
   listingUrgencyBookedShortRowContent(),
-]
+  ]
```

erichoracek:
Thinking about this further, Xcode indentation is actually **inconsistent** here, indenting the following **differently**:
```swift
var rowTypes = [
  .summary,
  ]
```
and
```swift
var rowTypes = [
  ""
]
```
And with trailing closure syntax, it indents the last trailing closure differently from the first trailing closure within chained function calls:
```swift
blah.func1 {
    print("")
  }
  .func2 {
    print("")    
}
```
Do we really want to live in a world where we have to memorize Xcode's inconsistencies and try to follow them when there's widely agreed upon alternatives that exist?

To me this hurts readability given that in almost all other languages the indentation of closing braces/literals matches the indentation of the corresponding opening brace/literal. If we have tooling to autocorrect this does the pain caused by having the linter occasionally autocorrect your pull request outweigh the cost of lowering readability and deviating from industry style norms throughout our codebase? And especially so since Xcode isn't even consistent about this.

bachand:
I don’t feel like I will need to memorize anything since Xcode's <kbd>^</kbd> + <kbd>I</kbd> indentation behavior will ensure that my code is in the format that Xcode wants. In a world where our tooling was more mature and autocorrection was truly stemless I may feel differently. But having our autocorrection tooling prevent me from pushing or fail my CI build when I did what Xcode wanted is enough of a productivity hit for me that I would prefer Xcode’s (albeit lackluster) world.

I feel like this is something we should file a Radar on as I can’t imagine this is the experience that Apple wants for its developers.

bachand:
Xcode’s chained closure indentation is crazy IMHO. I’ll be honest that I actively try to avoid this now by not chaining closures and instead making intermediate variables or otherwise refactoring my code. I’ve always been fairly happy with how my refactored code looks and it’s arguably more readable and easier to maintain since there is less happening on one logical line of code.

I’m sure people feel differently on this subject. I’m sharing my approach in case it’s useful to others.

gonzalonunez:
#25 is still open and has some valuable discussion around this as well. Should we close that PR and bring all discussion over to here?

rholmes:
@erichoracek If Xcode allowed indentation rules to be customized (which must be coming at some point right?) I doubt its current behavior would be my preference, but I don't find it to be a burden since it's applied via both autocomplete and ^ + I. I've been "using" this style for a long time and haven't felt the need to memorize it yet. 

On the other hand, I would feel at least some pressure to memorize areas where our style guide deviates from Xcode's default behavior to avoid reformatting. 

I think one implication of "don't fight Xcode" is that we should be able to easily write code in compliance with the style guide if we so choose. Rules that conflict with Xcode are a step in the opposite direction. The tenet is valuable precisely because it places more importance on avoiding these conflicts than on enforcing an "ideal" style. It defines a limit to the scope and (potentially negative) impact of the style guide, and it inevitably leads to accepting Xcode behavior that we may not like or wouldn't have chosen in the first place.

erichoracek:
Agreed that it would be ideal if Xcode supported customizable autocorrection so that we could customize the `^ + I` behavior to maximize readability. Until then, it seems like we should remove this rule given that folks are pretty strongly against increasing the number of lint failures we encounter, even if they have a positive impact on the readability of the codebase.

fdiaz:
## Resolution

#### Status: Approved ✅

#### Reasoning:

Even though the general consensus is that following Xcode indentation here hurts readability, not following creates a bigger friction for developers since they will have more lint failures after using Xcode default indentation or using `^ + I`. Given that this PR is approved and the rule removed from our style guide.

PR #84, opened 2019-03-12
Add custom_rules to whitelist

jingx23:
#### Summary

This will add custom_rules to the whitlist, otherwise they won´t get applied.

#### Reasoning
If using custom rules alongside a whitelist, make sure to add custom_rules as an item under whitelist_rules. See also [SwiftLint README](https://github.com/realm/SwiftLint/blob/master/README.md#defining-custom-rules)

#### Reviewers


fdiaz:
Thanks for the fix!

PR #86, opened 2019-04-14
Adjust Example

dpopov:
#### Summary

Updated Function Name in Example

#### Reasoning

Example should conform to previous rule: "Event-handling functions should be named like past-tense sentences"

#### Reviewers


fdiaz:
Makes sense! Thanks for contributing @dpopov and thanks for tagging me again @bachand ! I had totally missed this one

PR #88, opened 2019-07-09
Make rule details permanently visible

lewislbr:
#### Summary

Make rule details visible by default, eliminating the need to click 'Details' to see them.

#### Reasoning

I think it helps when scanning the document as sometimes is faster to identify the code at a glance than to read the rule description. Also, it reduces the interaction cost of having to click every rule details to see them. 

#### Reviewers


fdiaz:
Thanks for contributing @lewislbr! The style guide originally actually didn't have the 'Details' section. We added this because the style guide got too hard to read and skim through actually. When you have a wall of text it ends up being harder to find a specific rule. We wanted to optimize for finding specific rules instead of reading the whole guide from top to bottom (which is less useful for the most common use case - finding a specific rule).

Ideally, we'd have an "expand all" button, but we haven't figure out how to do that. Happy to reopen this if you think there's a better approach here.

lewislbr:
Thanks for the input, makes sense 👌

PR #89, opened 2019-08-06
Add style for upgrading weak self to strong reference

krunk4ever:
#### Summary

Starting in Swift 4.2 (released September 2018), `self` is allowed to be upgraded from a weak reference to strong reference: [Allow using optional binding to upgrade self from a weak to strong reference](https://github.com/apple/swift-evolution/blob/master/proposals/0079-upgrade-self-from-weak-to-strong.md)

#### Reasoning

I propose that we adopt this style for the same reasons it was accepted into Swift language.

* Avoid having to deal with `strongSelf`, `s`, `ss`
* Follows existing Swift convention of optional binding that reuses the name of the variable.

#### Reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,46 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='upgrade-self'></a>(<a href='#upgrade-self'>link</a>) **Bind to `self` when upgrading from a weak reference.** 
+
+  <details>
+
+  ```swift
+  //WRONG
+  class MyClass {
+
+    func request(completion: () -> Void) {
+      API.request() { [weak self] response in
+        guard let strongSelf = self else { return }
+        strongSelf.doSomething(strongSelf.property)
+        completion()
+      }
+    }
+    
+    func doSomething(nonOptionalParameter: SomeClass) {
+      // Processing and side effects
+    }
+  }
+
+  // RIGHT
+  class MyClass {
+
+    func request(completion: () -> Void) {
+      API.request() { [weak self] response in
+        guard let self = self else { return }
+        self.doSomething(self.property)
+        completion()
+      }
+    }
+
+    func doSomething(nonOptionalParameter: SomeClass) {
+      // Processing and side effects
+    }
+  }
+  ```
+
+  </details>
+
 * <a id='trailing-comma-array'></a>(<a href='#trailing-comma-array'>link</a>) **Add a trailing comma on the last element of a multi-line array.** [![SwiftFormat: trailingCommas](https://img.shields.io/badge/SwiftFormat-trailingCommas-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#trailingCommas)
 
   <details>
@@ -713,7 +753,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     func request(completion: () -> Void) {
       API.request() { [weak self] response in
-        if let strongSelf = self {
+        if let self = self {
           // Processing and side effects
         }
         completion()
@@ -726,8 +766,8 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     func request(completion: () -> Void) {
       API.request() { [weak self] response in
-        guard let strongSelf = self else { return }
-        strongSelf.doSomething(strongSelf.property)
+        guard let self = self else { return }
+        self.doSomething(self.property)
         completion()
       }
     }
--- a/README.md
+++ b/README.md
@@ -351,6 +351,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
       API.request() { [weak self] response in
         guard let strongSelf = self else { return }
         // Do work
+        completion()
       }
     }
   }
@@ -362,6 +363,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
       API.request() { [weak self] response in
         guard let self = self else { return }
         // Do work
+        completion()
       }
     }
   }
--- a/README.md
+++ b/README.md
@@ -350,13 +350,8 @@ _You can enable the following settings in Xcode by running [this script](resourc
     func request(completion: () -> Void) {
       API.request() { [weak self] response in
         guard let strongSelf = self else { return }
-        strongSelf.doSomething(strongSelf.property)
-        completion()
+        // Do work
       }
-    }
-    
-    func doSomething(nonOptionalParameter: SomeClass) {
-      // Processing and side effects
     }
   }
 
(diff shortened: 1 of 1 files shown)


bachand:
I would prefer recommending that we bind to `strongSelf` as I think it reduces cognitive load. When I see `self` in a closure, an alarm bell goes off in my head to be extra cautious about reviewing that code, whether it's mine or someone else's, for a retain cycle. With this guideline, `self` would now often be safe in closures. I like binding to `strongSelf` as it means that seeing `self` in a closure continues to be something we should carefully review.

With closures that are only a handful of lines this may not be a big deal since you can easily see at the closure declaration whether the programmer included `[weak self]`. But if there are long closures, you may not be able to easily see if `[weak self]` exists without having to expand the code more in GHE. I like it when code requires as little context as possible, and I think `strongSelf` helps achieve that.

fdiaz:
Thanks for contributing @krunk4ever !

Would it be possible to create a [lint rule in SwiftFormat](https://github.com/nicklockwood/SwiftFormat) for this? Ideally we'd like to have all of our rules automatically corrected.

No need to do this right now (although it'd be cool since I think the rest of the iOS community could benefit from this), but we'd require this before merging unless there's a compelling reason to not have this automatically corrected.

bachand:
I feel like it could be challenging to create a lint rule for this but I’ll be very happy if we can do it :)

krunk4ever:
@fdiaz Will take a look at creating a lint rule. 👍

krunk4ever:
Took me awhile, but I think I've come up with a pretty good regex. I created a custom rule in my swiftlint.yml file:

```
  upgrade_weak_self_to_strong_reference:
    name: "Upgrade self to strong reference"
    regex: '(let\s+((?!self)\S)*\s+=\s+self[^.?])'
    message: "Please reuse self as the variable name when upgrading a weak reference of self"
    severity: error
```

```
$ lib/swiftlint/0.26.0/bin/swiftlint lint --config swiftlint.yml --path "…/BookingGuestRequirementsViewController.swift" 
Loading configuration from 'swiftlint.yml'
Linting Swift files at path …/BookingGuestRequirementsViewController.swift
Linting 'BookingGuestRequirementsViewController.swift' (1/1)
…/BookingGuestRequirementsViewController.swift:233:15: error: Upgrade self to strong reference Violation: Please reuse self as the variable name when upgrading a weak reference of self (upgrade_weak_self_to_strong_reference)
…/BookingGuestRequirementsViewController.swift:251:15: error: Upgrade self to strong reference Violation: Please reuse self as the variable name when upgrading a weak reference of self (upgrade_weak_self_to_strong_reference)
…/BookingGuestRequirementsViewController.swift:283:15: error: Upgrade self to strong reference Violation: Please reuse self as the variable name when upgrading a weak reference of self (upgrade_weak_self_to_strong_reference)
…/BookingGuestRequirementsViewController.swift:305:15: error: Upgrade self to strong reference Violation: Please reuse self as the variable name when upgrading a weak reference of self (upgrade_weak_self_to_strong_reference)
Done linting! Found 4 violations, 4 serious in 1 file.
```

Similarly, you can replace the 1st `self` in my regex with `strongSelf` if we prefer to enforce that instead. Updating my custom rule resulted in:

```
$ lib/swiftlint/0.26.0/bin/swiftlint lint --config swiftlint.yml --path "…/BookingGuestRequirementsViewController.swift" 
Loading configuration from 'swiftlint.yml'
Linting Swift files at path …/BookingGuestRequirementsViewController.swift
Linting 'BookingGuestRequirementsViewController.swift' (1/1)
Done linting! Found 0 violations, 0 serious in 1 file.
```

Note: I truncated my path to make the logs more readable.

cc @bachand

krunk4ever:
These were test I did on https://rubular.com

Regex: `let\s+((?!self)\S)*\s+=\s+self[^.?]`

Test string:
```
// non-matching
if let self = self {
if let self = self,
if let x = self.property {
if let x = self?.property {
if let x = y, let self = self {
guard let self = self else {
let x = y a.b = self
let x y = self

// matching
let x = self {
let x = self,
let x = self as? NSObject
if let x = y, let z = self {
guard let x = self else {
```

bachand:
I was originally thinking we’d want to detect if we’re inside a closure but I’d be ok punting on that for now if we can’t think of meaningful false positives. I do think it’s fairly unlikely we’re assigning `self` to a variable in other cases.

I’ll defer to @fdiaz as to whether it’s ok we use SwiftLint for this instead of SwiftFormat.

bachand:
Does these work? Sorry I would test but I’m AFK.

```
guard let x = 1, y = self else {
```

```
guard
  let x = 1,
  y = self else
{
```

Also do you think we should allow for `var` as well as `let`. I think I’m comfortable with just `let` but I wanted to ask.

Thanks @krunk4ever 🙏

krunk4ever:
@bachand good call outs. i think i can refine the regex a bit more to include/exclude those extra cases. i wanted to use _swiftlint_ as a proof of concept since i was more familiar with it, but will look into using _SwiftFormat_

bachand:
Yeah SwiftLint is definitely more approachable. The big gain of SwiftFormat is the autocorrections. Since autocorrections would be hard here (we’d have to autocorrect many lines potentially) SwiftLint may be sufficient. I’d confirm with @fdiaz so that you don’t sink time into SwiftFormat if it’s not necessary. I do think it would be great to be fully on SwiftFormat one day. In my experience it’s harder to use for simple cases but overall more powerful, making complex cases possible.

fdiaz:
@krunk4ever we can use the new [strongifiedSelf](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#strongifiedself) rule from SwiftFormat here and merge this.

Let me know if you'd like to do that, if not I can do it for you and merge 😉

bachand:
Wow, neat new rule.
It seems like we should also lint against `strongSelf`, `ss`, and friends?

krunk4ever:
@fdiaz I think we can merge this and update SwiftFormat wit the new `strongifiedSelf` rule, but it looks like i'll only help identify when back ticks are used, i.e.

`let `self` = self ` => `let self = self`

and won't help identify when `ss` or `strongSelf` is used.

krunk4ever:
@fdiaz @bachand before we can merge this style guide update, a review is needed.

krunk4ever:
Posted this question to SwiftFormat: https://github.com/nicklockwood/SwiftFormat/issues/521

krunk4ever:
Have a PR out for auto-correct: https://github.com/nicklockwood/SwiftFormat/pull/523

fdiaz:
I'm actually ok with just merging this as is and trying to contribute back to SwiftFormat for more additions @airbnb/swift-styleguide-maintainers thoughts?

bachand:
Since the change to SwiftFormat that @krunk4ever has proposed will still require human intervention in some cases (eg to change variable names from `strongSelf` to `self` or address changes in the nullability of `self`) I’m ok merging this as-is too.

krunk4ever:
@bachand it doesn't look like i have write access to merge this PR. can you help me do it?

fdiaz:
@krunk4ever thank you for contributing and getting this to the finish line! 🎉

PR #90, opened 2019-08-15
Deemphasize `favor-constructors` and remove `use-modern-swift-extensions`

calda:
#### Summary

I propose that we update `favor-constructors` and remove `use-modern-swift-extensions`.

#### Reasoning

`CGRectMake`, `CGPointMake`, `CGRectZero`, `CGRectGetWidth` and many others are no longer available in Swift. (They were removed as long ago as Swift 3.0).

Unless there are other examples that still exist in the language, `use-modern-swift-extensions` no longer seems relevant. It's effectively compiler-enforced, so there's no need to suggest it 

`NSMakeRange` still exists, so `favor-constructors` should reference that example instead:

```swift
// Bad
let range = NSMakeRange(10, 5)

// Good
let range = NSRange(location: 10, length: 5)
```

#### Reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,36 +392,6 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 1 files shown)


bachand:
I think this raises a question: "to what versions of Swift does this style guide apply?"
That being said, I am supportive of this change.

bachand:
@fdiaz we have six upvotes after 10 days open. Should we merge this?

PR #91, opened 2019-10-03
Consider Markdown for Repository Language Determination

mattt:
#### Summary

Currently, GitHub incorrectly displays this repository as being a Shell project:

Screen Shot 2019-10-03 at 09 13 22

This PR adds a `.gitattributes` file so that GitHub will consider Markdown files when making its language determination for this repository:

Screen Shot 2019-10-03 at 09 12 54

#### Reasoning

This is a similar change to what I proposed on Swift Evolution (https://github.com/apple/swift-evolution/pull/1074), and much of the same rationale applies here as well. The basic gist is that correcting this categorization will help discoverability and encourage new contributions.  

#### Reviewers

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.md linguist-detectable=true
+*.md linguist-documentation=false


mattt:
Thanks for the quick merge, @fdiaz! Not sure why the language statistics haven't updated on the mainline repository yet, but 🤞 that'll get recalculated soon.

fdiaz:
@mattt I think it just took a bit to update, but it looks good now!

PR #92, opened 2019-10-25
Add "ID" example capitalize-acronyms rule, and make example more powerful

bachand:
#### Summary

- Removes `func isURLReachable(_:)`. IMHO this function was adding minimal educational value. The only difference between `func isURLReachable(_:)` and `isValidURL(_:)` is that "URL" is in the middle instead of the end. I don't think this is very debateable, and so I think we should simplify to focus on more interesting cases.
- The line of code `let urlValidator = URLValidator().isValidURL(/* some URL */)` doesn't make sense. We're creating a variable called `urlValidator` that is actually a `Bool` 🤦‍♂ The `urlValidator` variable now has semantic meaning.
- Adds examples that show how we recommend formatting "ID".

#### Reasoning

Apple's [`Identifiable` protocol](https://developer.apple.com/documentation/swift/identifiable) establishes that `ID` is [uppercase when used as a type](https://developer.apple.com/documentation/swift/identifiable/3285390-id) and [lowercase when used as a property](https://developer.apple.com/documentation/swift/identifiable/3285392-id). I believe that we should consider "ID" to be an acronym for the purposes of this rule.

#### Reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ _You can enable the following settings in Xcode by running [this script](resourc
     }
   }
 
+  let URLValidator = UrlValidator().isValidUrl(/* some URL */)
+
   // RIGHT
   class URLValidator {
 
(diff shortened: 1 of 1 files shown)


fdiaz:
Oops! Sorry for taking so long, I forgot about this one! Thanks for contributing :+1:

PR #95, opened 2020-01-30
Add omit-return rule

calda:
#### Summary

I propose we adopt a new `omit-return` rule:

> **Omit the `return` keyword when not required by the language.** 
> **[[SwiftFormat: redundantReturn]](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantReturn)**

```swift
// WRONG
["1", "2", "3"].compactMap { return Int($0) }

var size: CGSize {
  return CGSize(
    width: 100.0,
    height: 100.0)
}

func makeInfoAlert(message: String) -> UIAlertController {
  return UIAlertController(
    title: "ℹ️ Info",
    message: message,
    preferredStyle: .alert)
}

// RIGHT
["1", "2", "3"].compactMap { Int($0) }

var size: CGSize {
  CGSize(
    width: 100.0,
    height: 100.0)
}

func makeInfoAlert(message: String) -> UIAlertController {
  UIAlertController(
    title: "ℹ️ Info",
    message: message,
    preferredStyle: .alert)
}
```

#### Reasoning

- As of Swift 5.1 ([SE-255](https://github.com/apple/swift-evolution/blob/master/proposals/0255-omit-return.md)), the `return` keyword can now be omitted from single-expression functions and accessors.
- It isn't currently codified in the style guide, but omitting return is already strongly preferred for single-element closures
  - e.g. we would already tend to prefer `["1", "2", "3"].compactMap { Int($0) }` over `["1", "2", "3"].compactMap { return Int($0) }`.
- Our style guide prefers omitting keywords and types when permitted by the compiler:
  - Don't use `self` unless it's necessary for disambiguation or required by the language. (<a href='https://github.com/airbnb/swift#omit-self'>link</a>)
  - Don't include types where they can be easily inferred. (<a href='https://github.com/airbnb/swift#use-implicit-types'>link</a>)
  - Omit `Void` return types from function definitions. (<a href='https://github.com/airbnb/swift#omit-function-void-return'>link</a>)

#### Reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -1076,6 +1076,46 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='omit-single-expression-return'></a>(<a href='#omit-single-expression-return'>link</a>) **Omit the `return` keyword from single-expression functions, accessors, and closures.** [![SwiftFormat: redundantReturn](https://img.shields.io/badge/SwiftFormat-redundantReturn-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantReturn)
+
+  <details>
+  
+  ```swift
+  // WRONG
+  ["1", "2", "3"].compactMap { return Int($0) }
+  
+  var size: CGSize {
+    return CGSize(
+      width: 100.0,
+      height: 100.0)
+  }
+  
+  func makeInfoAlert(message: String) -> UIAlertController {
+    return UIAlertController(
+      title: "ℹ️ Info",
+      message: message,
+      preferredStyle: .alert)
+  }
+  
+  // RIGHT
+  ["1", "2", "3"].compactMap { Int($0) }
+  
+  var size: CGSize {
+    CGSize(
+      width: 100.0,
+      height: 100.0)
+  }
+  
+  func makeInfoAlert(message: String) -> UIAlertController {
+    UIAlertController(
+      title: "ℹ️ Info",
+      message: message,
+      preferredStyle: .alert)
+  }
+  ```
+  
+  </details>
+
 **[⬆ back to top](#table-of-contents)**
 
 ## File Organization
--- a/README.md
+++ b/README.md
@@ -1076,7 +1076,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='omit-single-expression-return'></a>(<a href='#omit-single-expression-return'>link</a>) **Omit the `return` keyword from single-expression functions, accessors, and closures.** [![SwiftFormat: redundantReturn](https://img.shields.io/badge/SwiftFormat-redundantReturn-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantReturn)
+* <a id='omit-single-expression-return'></a>(<a href='#omit-single-expression-return'>link</a>) **Omit the `return` keyword when not required by the language.** [![SwiftFormat: redundantReturn](https://img.shields.io/badge/SwiftFormat-redundantReturn-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantReturn)
 
   <details>
   
--- a/README.md
+++ b/README.md
@@ -1076,7 +1076,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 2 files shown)


erichoracek:
@calda This seems great and in line with the spirit of this styleguide. Do you imagine this would apply to return statements in single line closures as well? And do you know if the swiftlint rule omits return in single line closures? It appears that it does from the SwiftLint documentation.

calda:
@airbnb/swift-styleguide-maintainers Should new rules update `airbnb.swiftformat`, `swiftlint.yml`, or both?

bachand:
@calda preferably `airbnb.swiftformat`. No need for both.

README.md
```diff
@@ -1058,6 +1058,46 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+// WRONG
+["1", "2", "3"].compactMap { return Int($0) }
+
+var size: CGSize {
+  return CGSize(
+    width: 100.0,
+    height: 100.0)
+}
+
+func buildInfoAlert(message: String) -> UIAlertController {
```
erichoracek: nit: the Swift API design guidelines specify that:
> Begin names of factory methods with “make”, e.g. x.makeIterator().

https://swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage

So this would be `makeInfoAlert`

README.md
```diff
@@ -1076,6 +1076,46 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='omit-single-expression-return'></a>(<a href='#omit-single-expression-return'>link</a>) **Omit the `return` keyword from single-expression functions, accessors, and closures.** [![SwiftFormat: redundantReturn](https://img.shields.io/badge/SwiftFormat-redundantReturn-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantReturn)
```
fdiaz: Maybe:

```suggestion
* <a id='omit-single-expression-return'></a>(<a href='#omit-single-expression-return'>link</a>) **Omit the `return` keyword from single expressions when not required by the language.** [![SwiftFormat: redundantReturn](https://img.shields.io/badge/SwiftFormat-redundantReturn-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantReturn)
```

I don't think we need to be super specific about function, accessors and closures here

calda: @fdiaz This is a pedantic distinction, but in the [Swift grammar](https://docs.swift.org/swift-book/ReferenceManual/zzSummaryOfTheGrammar.html) the [`return`](https://docs.swift.org/swift-book/ReferenceManual/Statements.html#ID444) keyword isn't part of the [`expression`](https://docs.swift.org/swift-book/ReferenceManual/Expressions.html).

Here's the relevant grammar for a `function-body` (simplified to only show the relevant single-expression cases):
```
function-body → code-block

code-block → { statement }

statement → expression
statement → control-transfer-statement

control-transfer-statement → return-statement

return-statement → return expression 
```

In this grammar, `function-body` can either be `{ expression }` or `{ return expression }`. 

So within the Swift grammar, it's not strictly accurate to say that we can omit the `return` keyword _from an expression_. This grammar allows us to omit the `return` keyword from a single-expression function.

bachand: Could we simplify to this?

> **Omit the `return` keyword when not required by the language.**

calda: @bachand That sounds reasonable. There's existing precedent in the style guide for this sort of "required by the language" wording:

> Don't use self unless it's necessary for disambiguation or required by the language. ([link](https://github.com/airbnb/swift#omit-self))

bachand: I've been working on this style guide a long time @calda  😉

README.md
```diff
@@ -1076,6 +1076,46 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='omit-single-expression-return'></a>(<a href='#omit-single-expression-return'>link</a>) **Omit the `return` keyword when not required by the language.** [![SwiftFormat: redundantReturn](https://img.shields.io/badge/SwiftFormat-redundantReturn-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantReturn)
```
bachand: @calda should we update the link too, now that we've simplified the language?

calda: To just `#omit-return`?

bachand: That’s consistent with the `self` one, so yeah

PR #96, opened 2020-03-20
Update complex closure example

dcramps:
#### Summary

Hello old friends! I found a thing I don't like in the complex closure example, so I fixed it

#### Reasoning

As it is, the `RIGHT` example violates the unused closure parameter rule. Just adding `response` as a parameter to `doSomething` to make things more clear.

#### Reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -745,12 +745,12 @@ _You can enable the following settings in Xcode by running [this script](resourc
     func request(completion: () -> Void) {
       API.request() { [weak self] response in
         guard let self = self else { return }
-        self.doSomething(self.property)
+        self.doSomething(with: self.property, response: response)
         completion()
       }
     }
 
-    func doSomething(nonOptionalParameter: SomeClass) {
+    func doSomething(with nonOptionalParameter: SomeClass, response: SomeResponseClass) {
       // Processing and side effects
     }
   }


fdiaz:
Thanks!

README.md
```diff
@@ -745,12 +745,12 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
       API.request() { [weak self] response in
         guard let self = self else { return }
-        self.doSomething(self.property)
+        self.doSomething(with: self.property, response: response)
         completion()
       }
     }
 
-    func doSomething(nonOptionalParameter: SomeClass) {
+    func doSomething(with nonOptionalParameter: SomeClass, response: SomeResponseClass) {
```
dfed: Could we give the first parameter a more descriptive name?
```suggestion
    func doSomething(withNonOptionalParameter nonOptionalParameter: SomeClass, response: SomeResponseClass) {
```

Open to suggestions, but `with` isn't terrible descriptive as-is.

dcramps:
✌️ same to you @bachand !

README.md
dcramps: Yeah I considered this. Generally the class type would give you the context rather than a wordy label, so in reality `with` I think would be okay. Hard to be descriptive yet not overly verbose when your property is called `property` and the class is called `SomeClass`

dfed: Looks like [Swift's API guidelines](https://swift.org/documentation/api-design-guidelines/#promote-clear-usage) overrule me here. I got tripped up by our [hint at types](https://github.com/airbnb/swift#hint-at-types) rule, but our guide [does not override Swift's guidelines](https://github.com/airbnb/swift#guiding-tenets).

fdiaz: I actually tend to disagree. 

Based on [Promote Clear Usage](https://swift.org/documentation/api-design-guidelines/#promote-clear-usage )

```swift
public mutating func remove(at position: Index) -> Element
```

and [Parameter names](https://swift.org/documentation/api-design-guidelines/#parameter-names)

```swift
func move(from start: Point, to end: Point)
```

it's clear to me that we don't need to add more information to the parameter name when the type is strong.

This is not true when the type information is weak (e.g. a String or an Int). In that case it's better to move this information to the function name directly and then in the subsequent parameters.

See **Compensate for weak type information**

```diff
- func add(_ observer: NSObject, for keyPath: String)
+ func addObserver(_ observer: NSObject, forKeyPath path: String)
```

gonzalonunez:
@dcramps Brilliant profile picture 😂

Issue #97, opened 2020-06-19
Prefer AnyObject over class in protocol definitions


gsabran:
[This accepted Swift proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md#class-and-anyobject) that merges the concepts of `class` and `AnyObject` states that

> we suggest only keeping AnyObject around

We should update the guide to align with this evolution of Swift.

Popular linters such as Swift format and Swift lint already have rules for this:

https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#anyobjectprotocol

https://github.com/realm/SwiftLint/tree/master/Source/SwiftLintFramework/Rules/Lint

fdiaz:
this sounds good to me

bachand:
@gsabran could you put up a PR?

calda:
#99

PR #99, opened 2020-07-09
Prefer AnyObject over class in protocol definitions

gsabran:
#### Summary
https://github.com/airbnb/swift/issues/97

Use `: AnyObject` instead of `: class` in protocol definitions

#### Reasoning

[This accepted Swift proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md#class-and-anyobject) that merges the concepts of `class` and `AnyObject` states that

> we suggest only keeping AnyObject around

We should update the guide to align with this evolution of Swift.

Popular linters such as Swift format and Swift lint already have rules for this:

https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#anyobjectprotocol

https://github.com/realm/SwiftLint/tree/master/Source/SwiftLintFramework/Rules/Lint

### DIFF
--- a/README.md
+++ b/README.md
@@ -1119,6 +1119,11 @@ _You can enable the following settings in Xcode by running [this script](resourc
 * <a id='use-anyobject'></a>(<a href='#use-anyobject'>link</a>) **Use `AnyObject` instead of `class` in protocol definitions.** [![SwiftFormat: anyObjectProtocol](https://img.shields.io/badge/SwiftFormat-anyObjectProtocol-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#anyobjectprotocol)
 
   <details>
+  ### Why?
+  
+  [SE-0156](https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md]), which introduced support for using the `AnyObject` keyword as a protocol constraint, recommends preferring `AnyObject` over `class`:
+  
+  > This proposal merges the concepts of `class` and `AnyObject`, which now have the same meaning: they represent an existential for classes. To get rid of the duplication, we suggest only keeping `AnyObject` around. To reduce source-breakage to a minimum, `class` could be redefined as `typealias class = AnyObject` and give a deprecation warning on class for the first version of Swift this proposal is implemented in. Later, `class` could be removed in a subsequent version of Swift.
   
   ```swift
   // WRONG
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1116,6 +1116,20 @@ _You can enable the following settings in Xcode by running [this script](resourc
   
   </details>
 
+* <a id='use-anyobject'></a>(<a href='#use-anyobject'>link</a>) **Use `AnyObject` instead of `class` in protocol definitions.** [![SwiftFormat: anyObjectProtocol](https://img.shields.io/badge/SwiftFormat-anyObjectProtocol-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#anyobjectprotocol)
+
+  <details>
+
```
erichoracek, line 1122: Can we include a Why section that roughly matches your explanation from this PR?

calda: Perhaps: 
```suggestion
  ### Why?
  
  [SE-0156](https://github.com/apple/swift-evolution/blob/master/proposals/0156-subclass-existentials.md]), which introduced support for using the `AnyObject` keyword as a protocol constraint, recommends preferring `AnyObject` over `class`:
  
  > This proposal merges the concepts of `class` and `AnyObject`, which now have the same meaning: they represent an existential for classes. To get rid of the duplication, we suggest only keeping `AnyObject` around. To reduce source-breakage to a minimum, `class` could be redefined as `typealias class = AnyObject` and give a deprecation warning on class for the first version of Swift this proposal is implemented in. Later, `class` could be removed in a subsequent version of Swift.
  
```

calda:
Looks like there's lots of support for this new rule! 

@gsabran We can merge this once we [update `airbnb.swiftformat`](https://github.com/airbnb/swift/pull/99#pullrequestreview-445880494) and [add some additional rational](https://github.com/airbnb/swift/pull/99#discussion_r457739221) to the rule body.

resources/swiftlint.yml
```diff
@@ -1,4 +1,5 @@
 whitelist_rules:
+  - anyobject_protocol
```
calda, line 2: I don't think we need to update both `swiftlint.yml` and the `airbnb.swiftformat`. Since the rule links to the `anyObjectProtocol` SwiftFormat rule, I think we should just keep that one.

```suggestion
```

calda:
Thank you for your contribution @gsabran!

Issue #64, opened 2019-01-14
Computed properties and properties with property observers should appear at the end of the set of declarations of the same kind


fdiaz:
This rule is missing autocorrect. Following [this PR](https://github.com/airbnb/swift/pull/35).

---

* <a id='computed-properties-at-end'></a>(<a href='#computed-properties-at-end'>link</a>) **Computed properties and properties with property observers should appear at the end of the set of declarations of the same kind.** (e.g. instance properties.)

  <details>

  ```swift
  // WRONG
  var atmosphere: Atmosphere {
    didSet {
      print("oh my god, the atmosphere changed")
    }
  }
  var gravity: CGFloat

  // RIGHT
  var gravity: CGFloat
  var atmosphere: Atmosphere {
    didSet {
      print("oh my god, the atmosphere changed")
    }
  }
  ```

  </details>

calda:
Reintroduced in #102

Issue #62, opened 2019-01-14
Add empty lines between property declarations of different kinds


fdiaz:
This rule is missing autocorrect. Following [this PR](https://github.com/airbnb/swift/pull/35).

---

* <a id='newline-between-subsections'></a>(<a href='#newline-between-subsections'>link</a>) **Add empty lines between property declarations of different kinds.** (e.g. between static properties and instance properties.)

  <details>

  ```swift
  // WRONG
  static let gravityEarth: CGFloat = 9.8
  static let gravityMoon: CGFloat = 1.6
  var gravity: CGFloat

  // RIGHT
  static let gravityEarth: CGFloat = 9.8
  static let gravityMoon: CGFloat = 1.6

  var gravity: CGFloat
  ```

  </details>

calda:
Reintroduced in #102

Issue #60, opened 2019-01-14
Within each top-level section, place content in the following order


fdiaz:
This rule is missing autocorrect. Following [this PR](https://github.com/airbnb/swift/pull/35).

---

* <a id='subsection-organization'></a>(<a href='#subsection-organization'>link</a>) **Within each top-level section, place content in the following order.** This allows a new reader of your code to more easily find what he or she is looking for.
  * Constants (e.g. `static let gravity: CGFloat = 9.8`)
  * Static properties (e.g. `static let sharedInstance = Account()`)
  * Instance properties
  * Static methods
  * Class methods
  * Instance methods

calda:
Reintroduced in #102

Issue #56, opened 2019-01-14
Use // MARK: to separate the contents of a type definition


fdiaz:
This rule is missing autocorrect. Following [this PR](https://github.com/airbnb/swift/pull/35).

---

* <a id='marks-within-types'></a>(<a href='#marks-within-types'>link</a>) **Use `// MARK:` to separate the contents of a type definition into the sections listed below, in order.** All type definitions should be divided up in this consistent way, allowing a new reader of your code to easily jump to what he or she is interested in. [![SwiftLint: mark](https://img.shields.io/badge/SwiftLint-mark-008489.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#mark)
  * `// MARK: Lifecycle` for `init` and `deinit` methods.
  * `// MARK: Open` for `open` properties and methods.
  * `// MARK: Public` for `public` properties and methods.
  * `// MARK: Internal` for `internal` properties and methods.
  * `// MARK: Fileprivate` for `fileprivate` properties and methods.
  * `// MARK: Private` for `private` properties and methods.
  * If the type in question is an enum, its cases should go above the first `// MARK:`.
  * If there are typealiases, they should go above the first `// MARK:`.
  * Do not subdivide each of these sections into subsections, as it makes the method dropdown more cluttered and therefore less useful. Instead, group methods by functionality and use smart naming to make clear which methods are related. If there are enough methods that sub-sections seem necessary, consider refactoring your code into multiple types.
  * If the type in question is a simple value type, it is OK to omit the `// MARK:`s, as it would hurt legibility.

calda:
Reintroduced in #102

PR #102, opened 2020-08-28
Reintroduce old file organization rules

calda:
#### Summary

This is a PR reintroduces several of our old file organization rules:
 - [Use // MARK: to separate the contents of a type definition](https://github.com/airbnb/swift/issues/56)
- [Within each top-level section, place content in the following order](https://github.com/airbnb/swift/issues/60)
 - [Computed properties and properties with property observers should appear at the end of the set of declarations of the same kind ](https://github.com/airbnb/swift/issues/64)
 - [Add empty lines between property declarations of different kinds](https://github.com/airbnb/swift/issues/62)

We removed these rules because they were missing autocorrect, but they are now supported by SwiftFormat `0.46`'s new `organizeDeclarations` rule (nicklockwood/SwiftFormat#701).

I made a few small tweaks to the previous rules in [`b13000b`](https://github.com/airbnb/swift/pull/102/commits/b13000bd877191c698f0eb1aed25affcde7ed2e6). Discussion inline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1374,12 +1374,12 @@ _You can enable the following settings in Xcode by running [this script](resourc
   * `// MARK: Fileprivate` for `fileprivate` properties and methods.
   * `// MARK: Private` for `private` properties and methods.
   * If the type in question is an enum, its cases should go above the first `// MARK:`.
-  * If there are typealiases or nested types, they should go above the first `// MARK:`.
   * Do not subdivide each of these sections into subsections, as it makes the method dropdown more cluttered and therefore less useful. Instead, group methods by functionality and use smart naming to make clear which methods are related. If there are enough methods that sub-sections seem necessary, consider refactoring your code into multiple types.
   * If all of a type's definitions belong to the same category (e.g. the type only consists of `internal` properties), it is OK to omit the `// MARK:`s.
   * If the type in question is a simple value type (e.g. fewer than 20 lines), it is OK to omit the `// MARK:`s, as it would hurt legibility.
 
 * <a id='subsection-organization'></a>(<a href='#subsection-organization'>link</a>) **Within each top-level section, place content in the following order.** This allows a new reader of your code to more easily find what they are looking for. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
+  * Nested types
   * Static properties
   * Class properties
   * Instance properties
--- a/README.md
+++ b/README.md
@@ -1366,6 +1366,69 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 * <a id='newline-at-eof'></a>(<a href='#newline-at-eof'>link</a>) **Files should end in a newline.** [![SwiftLint: trailing_newline](https://img.shields.io/badge/SwiftLint-trailing__newline-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-newline)
 
+* <a id='marks-within-types'></a>(<a href='#marks-within-types'>link</a>) **Use `// MARK:` to separate the contents of a type definition into the sections listed below, in order.** All type definitions should be divided up in this consistent way, allowing a new reader of your code to easily jump to what they are interested in. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
+  * `// MARK: Lifecycle` for `init` and `deinit` methods.
+  * `// MARK: Open` for `open` properties and methods.
+  * `// MARK: Public` for `public` properties and methods.
+  * `// MARK: Internal` for `internal` properties and methods.
+  * `// MARK: Fileprivate` for `fileprivate` properties and methods.
+  * `// MARK: Private` for `private` properties and methods.
+  * If the type in question is an enum, its cases should go above the first `// MARK:`.
+  * If there are typealiases, they should go above the first `// MARK:`.
+  * Do not subdivide each of these sections into subsections, as it makes the method dropdown more cluttered and therefore less useful. Instead, group methods by functionality and use smart naming to make clear which methods are related. If there are enough methods that sub-sections seem necessary, consider refactoring your code into multiple types.
+  * If the type in question is a simple value type, it is OK to omit the `// MARK:`s, as it would hurt legibility.
+
+* <a id='subsection-organization'></a>(<a href='#subsection-organization'>link</a>) **Within each top-level section, place content in the following order.** This allows a new reader of your code to more easily find what they are looking for. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
+  * Constants (e.g. `static let gravity: CGFloat = 9.8`)
+  * Static properties (e.g. `static let sharedInstance = Account()`)
+  * Instance properties
+  * Static methods
+  * Class methods
+  * Instance methods
+  
+* <a id='newline-between-subsections'></a>(<a href='#newline-between-subsections'>link</a>) **Add empty lines between property declarations of different kinds.** (e.g. between static properties and instance properties.) [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
+
+  <details>
+
+  ```swift
+  // WRONG
+  static let gravityEarth: CGFloat = 9.8
+  static let gravityMoon: CGFloat = 1.6
+  var gravity: CGFloat
+
+  // RIGHT
+  static let gravityEarth: CGFloat = 9.8
+  static let gravityMoon: CGFloat = 1.6
+
+  var gravity: CGFloat
+  ```
+
+  </details>
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1374,13 +1374,14 @@ _You can enable the following settings in Xcode by running [this script](resourc
   * `// MARK: Fileprivate` for `fileprivate` properties and methods.
   * `// MARK: Private` for `private` properties and methods.
   * If the type in question is an enum, its cases should go above the first `// MARK:`.
-  * If there are typealiases, they should go above the first `// MARK:`.
+  * If there are typealiases or nested types, they should go above the first `// MARK:`.
```
calda, line 1377: This rule previously did not specify where we should place nested types.

I think they we should place them at the top, above the first `// MARK:`. This matches the existing policy for `typealias` declarations.

README.md
```diff
@@ -1374,13 +1374,14 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
   * Do not subdivide each of these sections into subsections, as it makes the method dropdown more cluttered and therefore less useful. Instead, group methods by functionality and use smart naming to make clear which methods are related. If there are enough methods that sub-sections seem necessary, consider refactoring your code into multiple types.
-  * If the type in question is a simple value type, it is OK to omit the `// MARK:`s, as it would hurt legibility.
+  * If all of a type's definitions belong to the same category (e.g. the type only consists of `internal` properties), it is OK to omit the `// MARK:`s.
```
calda, line 1379: This is how the `organizeDeclarations` rule behaves, so I think we should capture it in the rule description.

README.md
```diff
@@ -1374,13 +1374,14 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  * If the type in question is a simple value type (e.g. fewer than 20 lines), it is OK to omit the `// MARK:`s, as it would hurt legibility.
```
calda, line 1380: Now that this is formatted automatically, we need to explicitly specify what counts as a "simple value type".

The `organizeDeclarations` rule supports specifying a minimum line count for when we should auto-organize a type. I think 20 lines could be a good value here, but open to other suggestions.

```
--enumthreshold 20 # organizeDeclarations
```

README.md
```diff
@@ -1374,13 +1374,14 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
 
 * <a id='subsection-organization'></a>(<a href='#subsection-organization'>link</a>) **Within each top-level section, place content in the following order.** This allows a new reader of your code to more easily find what they are looking for. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
-  * Constants (e.g. `static let gravity: CGFloat = 9.8`)
-  * Static properties (e.g. `static let sharedInstance = Account()`)
+  * Static properties
+  * Class properties
```
calda, line 1384: `class` properties exist in Swift, so we should specify where they should be placed

README.md
```diff
@@ -1374,13 +1374,14 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
-  * If the type in question is a simple value type, it is OK to omit the `// MARK:`s, as it would hurt legibility.
+  * If all of a type's definitions belong to the same category (e.g. the type only consists of `internal` properties), it is OK to omit the `// MARK:`s.
+  * If the type in question is a simple value type (e.g. fewer than 20 lines), it is OK to omit the `// MARK:`s, as it would hurt legibility.
 
 * <a id='subsection-organization'></a>(<a href='#subsection-organization'>link</a>) **Within each top-level section, place content in the following order.** This allows a new reader of your code to more easily find what they are looking for. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
-  * Constants (e.g. `static let gravity: CGFloat = 9.8`)
-  * Static properties (e.g. `static let sharedInstance = Account()`)
+  * Static properties
```
calda, line 1383: The `organizeDeclarations` rule doesn't currently support the distinction between "constants" and "properties".

We could leave this as a suggestion, but we don't currently have the capability to enforce this in CI.

bachand:
@kompfner do you believe it?

README.md
bachand: This makes sense to me.

README.md
bachand: I'm OK with this.

README.md
bachand: This seems fine to me.

README.md
```diff
@@ -1366,6 +1366,70 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 * <a id='newline-at-eof'></a>(<a href='#newline-at-eof'>link</a>) **Files should end in a newline.** [![SwiftLint: trailing_newline](https://img.shields.io/badge/SwiftLint-trailing__newline-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-newline)
 
+* <a id='marks-within-types'></a>(<a href='#marks-within-types'>link</a>) **Use `// MARK:` to separate the contents of a type definition into the sections listed below, in order.** All type definitions should be divided up in this consistent way, allowing a new reader of your code to easily jump to what they are interested in. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
+  * `// MARK: Lifecycle` for `init` and `deinit` methods.
```
bachand, line 1370: @soopa if you can believe it, we now lint and autocorrect this in our code base. I still remember discussing this convention while working on Tungsten together.

resources/airbnb.swiftformat
```diff
@@ -11,7 +11,10 @@
 --closingparen same-line # wrapArguments
 --funcattributes prev-line # wrapAttributes
 --typeattributes prev-line # wrapAttributes
+--beforemarks typealias,class,struct,enum,protocol # organizeDeclarations
+--structthreshold 20 # organizeDeclarations
+--enumthreshold 20 # organizeDeclarations
```
bachand, line 16: @calda do we need a `--classthreshold` too?

calda:
> I'm assuming that the first commit brings back the old rules verbatim (with the SwiftFormat badges added) and that the only changes to the old rules are in b13000b.

This is correct!

> If that is true, you have my strong LGTM. This PR really warms my heart. Thank you.

:belo-heart:

resources/airbnb.swiftformat
calda: > If the type in question is a simple value type (e.g. fewer than 20 lines), it is OK to omit the `// MARK:`s, as it would hurt legibility.

My interpretation of the "simple value type" phrasing suggests this exception applies to `struct` and `enum` declarations (value types) and not `class` declarations (reference types). 

Happy to add `--classthreshold 20` as well if you think that is a more correct interpretation of the old rule (perhaps we could reword it as well?).

README.md
erichoracek: How does this handle types that have a specific access control level? I'd be concerned if nested private types were grouped with public types

calda: This places all nested types and typealiases (regardless of access control) above the first `// MARK:`. 

I personally think this is reasonable, but I can see the argument against putting `private` declarations at the top of a type.

calda: [`d0d660e`](https://github.com/airbnb/swift/pull/102/commits/d0d660e644569894d67d5787d414f7c18f074b38) removes this exception in favor of grouping nested types at the top of their respective access-control category.

README.md
```diff
@@ -1374,12 +1374,12 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
   * Do not subdivide each of these sections into subsections, as it makes the method dropdown more cluttered and therefore less useful. Instead, group methods by functionality and use smart naming to make clear which methods are related. If there are enough methods that sub-sections seem necessary, consider refactoring your code into multiple types.
   * If all of a type's definitions belong to the same category (e.g. the type only consists of `internal` properties), it is OK to omit the `// MARK:`s.
   * If the type in question is a simple value type (e.g. fewer than 20 lines), it is OK to omit the `// MARK:`s, as it would hurt legibility.
 
 * <a id='subsection-organization'></a>(<a href='#subsection-organization'>link</a>) **Within each top-level section, place content in the following order.** This allows a new reader of your code to more easily find what they are looking for. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
+  * Nested types
```
calda, line 1382: https://github.com/airbnb/swift/pull/102#discussion_r479437997

bachand, line 1382: @calda do you see value in clarifying that tyepaliases go here too (assuming that is correct)?

bachand: Should be a simple follow up PR. I think the clarity could be nice.

README.md
bachand: I think that's a good change.

resources/airbnb.swiftformat
bachand: Ah I misread the old rule. I agree that this is the correct interpretation. Thanks for being so thorough!

README.md
calda: Right now the `organizeDeclarations` rule will place `typealias` declarations with the instance properties (https://github.com/nicklockwood/SwiftFormat/blob/master/Sources/Rules.swift#L5072). 

I agree that we should specify exactly where they should go. If we want to place them somewhere else, I can make a follow-up SwiftFormat enhancement.

bachand: @calda it seems like it would be logic to group them with nested types. I don't feel strongly that they are positioned _before_ or _after_ nested types that are defined inline (rather than typealiased). This seems like it should be _valid_:


```swift
class Foo {

  // MARK: Public

  public enum MyEnum { }

  public typealias Bar = String

  public struct YourStruct { }

  // MARK: Internal
  
  var value = 1
}
```

bachand:
Also @calda don't forget to take your victory lap and close out those issues!

PR #103, opened 2020-08-31
Explicitly mention typealiases in the subsection organization rule

calda:
Follow-up to https://github.com/airbnb/swift/pull/102#discussion_r479558324.

Starting in SwiftFormat `0.46.1` (nicklockwood/SwiftFormat#724), `typealias` declarations are grouped with nested types. We should state this explicitly in the rule description.


bachand:
Nice! I wonder if we should add a note to our style guide which mentions which version of SwiftFormat we are using.

fdiaz:
@bachand I like that! Maybe as a comment on top of https://github.com/airbnb/swift/blob/master/resources/airbnb.swiftformat for now?

PR #105, opened 2020-09-25
Specify the access control for each declaration in an extension individually

calda:
#### Summary

This PR proposes a new rule that we should specify the access control for each declaration in an extension individually:

  ```swift
  // WRONG
  public extension Foo {
    // This declaration doesn't have an explicit access control level.
    // In all other scopes, this would be an internal function,
    // but because this is in a public extension, it's actually a public function.
    func bar() {}
  }

  // WRONG
  private extension Foo {
    func bar() {}
  }

  // RIGHT
  extension Foo {
    // It is immediately obvious that this is a public function,
    // even if the start of the `extension Foo` scope is off-screen.
    public func bar() {}
  }

  // RIGHT
  extension Foo {
    // Recall that a private extension actually has fileprivate semantics,
    // so a declaration in a private extension is fileprivate by default.
    fileprivate func bar() {}
  }
  ```

This PR also proposes extending the [mark-within-types](https://github.com/airbnb/swift#marks-within-types) rule to apply to extensions

#### Reasoning

The next version of SwiftFormat will include a new `extensionAccessControl` rule that supports this behavior (nicklockwood/SwiftFormat#742) and should include support for organizing the body of `extension` declarations using the `organizeDeclarations` rule (nicklockwood/SwiftFormat#742).

I think we should adopt the new rule because:
 - Specifying the access control on the declaration itself helps engineers more quickly determine the access control level of an individual declaration.
     - An `extension` is the only scope where a declaration does not necessarily default to being `internal`. 
     - Because of that, you don't know the access control level of a given declaration unless you also know the access control level of the extension declaration itself (which may not be on screen).
     - Removing this exception simplifies the mental model of access control.
 - Adopting this rule lets us extend the [mark-within-types](https://github.com/airbnb/swift#marks-within-types) rule to apply to extensions.
     - [mark-within-types](https://github.com/airbnb/swift#marks-within-types) should apply to all declaration scopes that support mixed access control (i.e. types and extensions, but not protocols).
     - It doesn't make sense to organize the body of an extension into access control categories when the individual declarations inherit an access control level from the extension itself. For example:
       ```swift
       // Unusual, and different from how mark-within-types is applied to type bodies
       public extension Foo {
      
         // MARK: Public

         func bar() {}

         // MARK: Internal

         internal func baaz() {}
        
       }

       // Better, and matches how mark-within-types is applied to type bodies 
       extension Foo {
      
         // MARK: Public

         public func bar() {}

         // MARK: Internal

         func baaz() {}
        
       }
       ```


README.md
```diff
@@ -1296,6 +1296,45 @@ _You can enable the following settings in Xcode by running [this script](resourc
   
   </details>
 
+* <a id='extension-access-control'></a>(<a href='#extension-access-control'>link</a>) **Specify the access control for each declaration in an extension individually.** [![SwiftFormat: extensionAccessControl](https://img.shields.io/badge/SwiftFormat-anyObjectProtocol-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md##extensionaccesscontrol)
```
bachand, line 1299: Is `anyObjectProtocol` a typo?

bachand, line 1299:
```suggestion
* <a id='extension-access-control'></a>(<a href='#extension-access-control'>link</a>) **Specify the access control for each declaration in an extension individually.** [![SwiftFormat: extensionAccessControl](https://img.shields.io/badge/SwiftFormat-anyObjectProtocol-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#extensionaccesscontrol)
```

?

README.md
```diff
@@ -1296,6 +1296,45 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  <details>
+
+  #### Why?
+
+  Specifying the access control on the declaration itself helps engineers more quickly determine the access control level of an individual declaration.
```
bachand, line 1305: Can we mention something about not having to look at the access control of the type you are extending? I think pointing out this locality issue would be great.

README.md
```diff
@@ -1296,6 +1296,45 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  // WRONG
+  private extension Foo {
+    func bar() {}
+  }
+
+  // RIGHT
+  extension Foo {
+    // It is immediately obvious that this is a public function,
+    // even if the start of the `extension Foo` scope is off-screen.
+    public func bar() {}
```
bachand, line 1325:
```suggestion
    public func bar() { }
```

bachand: Nit: not in the guide AFAIK but I think this style is nice

README.md
```diff
@@ -1296,6 +1296,45 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  }
+
+  // RIGHT
+  extension Foo {
+    // Recall that a private extension actually has fileprivate semantics,
+    // so a declaration in a private extension is fileprivate by default.
+    fileprivate func bar() {}
```
bachand, line 1332:
```suggestion
    fileprivate func bar() { }
```

README.md
```diff
@@ -1296,6 +1296,45 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  <details>
+
+  #### Why?
+
+  Specifying the access control on the declaration itself helps engineers more quickly determine the access control level of an individual declaration.
+
+  ```swift
+  // WRONG
+  public extension Foo {
```
bachand, line 1309: Totally optional: what do you think of coming up with more fun names?

README.md
```diff
@@ -1296,6 +1296,45 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  // RIGHT
+  extension Foo {
+    // It is immediately obvious that this is a public function,
+    // even if the start of the `extension Foo` scope is off-screen.
+    public func bar() {}
+  }
+
+  // RIGHT
+  extension Foo {
+    // Recall that a private extension actually has fileprivate semantics,
```
bachand, line 1330: I think this comment is a bit hard to follow since we're extending `Foo` a bunch of times. I think you are referring to the case where we write a `private` extension on `Foo`. Maybe we can make each paired example extend a unique type for clarity? And maybe have different method names?

It's nice when you could put this code into a playground and it would compile, though not mandatory.

README.md
```diff
@@ -1366,7 +1405,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
 
-* <a id='marks-within-types'></a>(<a href='#marks-within-types'>link</a>) **Use `// MARK:` to separate the contents of a type definition into the sections listed below, in order.** All type definitions should be divided up in this consistent way, allowing a new reader of your code to easily jump to what they are interested in. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
+* <a id='marks-within-types'></a>(<a href='#marks-within-types'>link</a>) **Use `// MARK:` to separate the contents of type definitions and extensions into the sections listed below, in order.** All type definitions and extensions should be divided up in this consistent way, allowing a new reader of your code to easily jump to what they are interested in. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
```
bachand, line 1408:
```suggestion
* <a id='marks-within-types'></a>(<a href='#marks-within-types'>link</a>) **Use `// MARK:` to separate the contents of type definitions and extensions into the sections listed below, in order.** All type definitions and extensions should be divided up in this consistent way, allowing a reader of your code to easily jump to what they are interested in. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
```

I know this was preexisting but what do you think of removing "new"? Seems like it applies to all readers.

README.md
bachand, line 1299: Will this rule automatically move access control from before the extension to each member?

bachand (commented):
Nice!

README.md
calda: bad copy-paste, thanks for catching this!

README.md
calda: Good suggestion! Agreed.

README.md
calda: Sure, I could use the `Spaceship` example that we use elsewhere in the style guide.

README.md
calda: Yes! Here's an example of what the SwiftFormat rule will do:

```diff
- public extension Foo {
-     func bar() {}
-     func baaz() {}
-     internal func quux() {}
  }

+ extension Foo {
+     public func bar() {}
+     public func baaz() {}
+     func quux() {}
  }
```

README.md
calda: What do you mean? Could you share an example?

README.md
calda: Updated: 

  ```swift
  extension Universe {
    public func generateGalaxy() { }
  }

  extension Spaceship {
    fileprivate func enableHyperdrive() { }
  }
  ```

README.md
bachand: Yeah definitely. I meant encoding some of this in the guide itself.

Screen Shot 2020-09-25 at 8 01 57 PM

bachand: At some point we may also want to go through and start saying "member declaration" for clarity, but I like that we're being consistent with the existing guide.

README.md
```diff
@@ -1306,30 +1306,30 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   ```swift
   // WRONG
-  public extension Foo {
```
bachand: I like `Foo` and `Bar` and don't intend to eliminate them from my "vocabulary" (e.g. see [this post](https://stackoverflow.com/questions/64072242/why-cant-i-pass-an-object-conforming-to-a-protocol-to-a-function-that-expects-a) I made today) but I think more human examples will make this guide accessible to a broader audience 👍

README.md
bachand: Though I see we cover it below, so I actually think it's fine 👍 
Screen Shot 2020-09-25 at 8 10 01 PM

bachand (approved):

calda:
I'll merge this PR once SwiftFormat `0.47` (with the new `extensionAccessControl` rule)  is released.

fdiaz (approved):

Issue #58, opened 2019-01-14
Each protocol conformance implementation should occur in dedicated type extension within the same file as the type


fdiaz:
This rule is missing autocorrect. Following [this PR](https://github.com/airbnb/swift/pull/35).

---

* <a id='extensions-for-protocol-conformance'></a>(<a href='#extensions-for-protocol-conformance'>link</a>)
 **Each protocol conformance implementation should occur in dedicated type extension within the same file as the type.** This extension should be marked with `// MARK: ProtocolName`, and should contain nothing more than the methods or properties required to conform to the protocol. As a result, no `// MARK:`s are needed for defining subsections.

calda:
Closing since this has been open for three years. I like this one personally but its probably intractable to autocomplete -- it would require having full type information available to the formatter tool (not only of the source code files being formatted, but all of their dependencies and the iOS SDKs etc).

PR #106, opened 2020-09-25
Remove "Prefer putting constants in the top level of a file if they are private" rule

calda:
#### Summary

This PR proposes removing this rule from our style guide:

> **Prefer putting constants in the top level of a file if they are `private`.** If they are `public` or `internal`, define them as static properties, for namespacing purposes.
> 
>   ```swift
>   private let privateValue = "secret"
>
>   public class MyClass {
>
>     public static let publicValue = "something"
>
>     func doSomething() {
>       print(privateValue)
>       print(MyClass.publicValue)
>     }
>
>   }
>   ```

Instead, I think we should recommend placing these constants in `enum` namespaces (as an extension to the existing [namespace-using-enums](https://github.com/airbnb/swift#namespace-using-enums) rule).

#### Reasoning

1. I don't think we should be suggesting that folks create top-level declarations. This feels at-odds with the [Avoid creating non-namespaced global constants and functions](https://github.com/airbnb/swift#namespace-using-enums) rule.
2. This is a really old rule (that was [added](https://github.com/airbnb/swift/commit/c93ee5aae718d25401e001e78308bef73a4b5ac9#diff-9ba7d4d50c31dd7166688703e242142dR9) in 2016!). I don't think this rule represents idiomatic Swift anymore, now that the language has evolved and `enum` namespaces have become much more common.

### DIFF
--- a/README.md
+++ b/README.md
@@ -959,7 +959,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Prefer using caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.**
+* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.**
   * Feel free to nest namespaces where it adds clarity.
   * Avoid creating non-namespaced global constants and functions.
   * `private` globals are permitted, since they are scoped to a single file and do not pollute the global namespace. Consider placing private globals in an `enum` namespace to match the guidelines for other declaration types.
--- a/README.md
+++ b/README.md
@@ -960,8 +960,8 @@ _You can enable the following settings in Xcode by running [this script](resourc
   </details>
 
 * <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.**
-  * Feel free to nest namespaces where it adds clarity.
   * Avoid creating non-namespaced global constants and functions.
+  * Feel free to nest namespaces where it adds clarity.
   * `private` globals are permitted, since they are scoped to a single file and do not pollute the global namespace. Consider placing private globals in an `enum` namespace to match the guidelines for other declaration types.
 
   <details>
--- a/README.md
+++ b/README.md
@@ -959,7 +959,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
+* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
 
   <details>
 
--- a/README.md
+++ b/README.md
@@ -962,7 +962,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 * <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Prefer using caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.**
   * Feel free to nest namespaces where it adds clarity.
   * Avoid creating non-namespaced global constants and functions.
-  * `private` globals are permitted, since they are scoped to a single file and do not pollute the global namespace. Consider placing private globals in an enum namespace to match the guidelines for other declaration types.
+  * `private` globals are permitted, since they are scoped to a single file and do not pollute the global namespace. Consider placing private globals in an `enum` namespace to match the guidelines for other declaration types.
 
   <details>
 
--- a/README.md
+++ b/README.md
@@ -959,7 +959,10 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
+* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Prefer using caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.**
+  * Feel free to nest namespaces where it adds clarity.
+  * Avoid creating non-namespaced global constants and functions.
+  * `private` globals are permitted, since they are scoped to a single file and do not pollute the global namespace. Consider placing private globals in an enum namespace to match the guidelines for other declaration types.
 
   <details>
 
--- a/README.md
+++ b/README.md
@@ -959,26 +959,6 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='private-constants'></a>(<a href='#private-constants'>link</a>) **Prefer putting constants in the top level of a file if they are `private`.** If they are `public` or `internal`, define them as static properties, for namespacing purposes.
-
-  <details>
-
-  ```swift
-  private let privateValue = "secret"
-
-  public class MyClass {
-
-    public static let publicValue = "something"
-
-    func doSomething() {
-      print(privateValue)
-      print(MyClass.publicValue)
-    }
-  }
-  ```
-
-  </details>
-
 * <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
 
   <details>


README.md
```diff
@@ -959,26 +959,6 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
```
bachand, line 960: >  I don't think we should be suggesting that folks create top-level declarations. This feels at-odds with the [Avoid creating non-namespaced global constants and functions](https://github.com/airbnb/swift#namespace-using-enums) rule.

I believe that this rule is synergistic with the [Avoid creating non-namespaced global constants and functions](https://github.com/airbnb/swift#namespace-using-enums) rule that you referenced.

You will notice that this rule explicitly calls out `public` and `internal` in its headline: "Use caseless enums for organizing `public` or `internal` constants and functions into namespaces". I think the "Avoid creating non-namespaced global constants and functions" needs to be read acknowledging that it's specific to `public` and `internal` constants and fucntions.

In the commit to which you linked, the rule at the time also mentions namespacing public constants: "If they are public, put the constant as a static property, so we get nice namespaces." I think this is further evidence that they are synergistic. Eventually that was made it owns rule. I would have to dig deeper into the history to remember it all.

At least one member of the team felt like it was overkill to create a caseless `enum` for private constants. I generally do try to avoid putting functions and properties in the global namespace. But it is a feature of the language. I think allowing `private` functions and properties in the global namespace is a good compromise, since they are isolated to a single file.

bachand, line 960: > This is a really old rule (that was [added](https://github.com/airbnb/swift/commit/c93ee5aae718d25401e001e78308bef73a4b5ac9#diff-9ba7d4d50c31dd7166688703e242142dR9) in 2016!). I don't think this rule represents idiomatic Swift anymore, now that the language has evolved and `enum` namespaces have become much more common.

What new information has come to light between now and 2016 that indicates that this is pattern is no idiomatic to Swift?

bachand, line 960: > This rule changes the format of code, but is not reformatted automatically.

Yes, when we did a pass to remove rules that are not automatically corrected we audited all rules to determine which could be easily autocorrected. We must have determined that this and it's [partner rule](https://github.com/airbnb/swift#namespace-using-enums) were not reasonably autocorrectable.

Tooling is getting better and maybe this is more feasible now. But actually updating all source code to organize constants like this would be fairly complex and we must have determined it was not really feasible.

bachand: @fdiaz may have thoughts

bachand (commented):
I added my comments to threads to allow for easier discussion.

README.md
bachand: cc: @jqsilver who wrote the original commit

calda: My general feeling is that the community has moved away from using unnamespaced globals over the past few years (which, anecdotally, seemed a lot more common in 2016 than it seems today in 2020).

calda: > But actually updating all source code to organize constants like this would be fairly complex and we must have 
determined it was not really feasible.

This seems approximately as challenging to autocorrect as the [mark-within-types](https://github.com/airbnb/swift#marks-within-types) rule, which we did remove from the Style Guide until it had autocorrect available. 

Are there differences you have in mind that would make this rule more challenging to autocorrect than [mark-within-types](https://github.com/airbnb/swift#marks-within-types)?

jqsilver:
I think that @bachand is right that private top-level constants don't violate our namespacing rules. The visibility of private top-level constants are limited to the file. However, I do think that my experience has been that it's better to use types to scope constants (and methods) instead of using the per-file visibility rules.

README.md
bachand: Let's look internally to see if we can find any documentation on how we went about deciding which rules to remove. That being said, it seems like this line of reasoning would also indicate that we should remove [namespace-using-enums](https://github.com/airbnb/swift#namespace-using-enums). Would you agree?

bachand, line 960: @calda if we remove this rule we have no guidance on `private` constants. Could we explore a replacement to cover this case so that our guide covers constants with any access control?

calda: In contrast to this rule, I think [namespace-using-enums](https://github.com/airbnb/swift#namespace-using-enums) actually _is_ impossible to autocorrect. 

 - For this global `private` constant rule, it would be reasonably doable to simply move a `private static` declaration to the top-level scope. This wouldn't break any accessors in the file, except ones using `self` (which could be patched by the linter).

 - In contrast, for [namespace-using-enums](https://github.com/airbnb/swift#namespace-using-enums), the linter wouldn't know what name to use for the `enum` namespace. Additionally, since these symbols are `public`, changing their name / namespace is a source-breaking change across the entire codebase (all of our existing linter tools operate on the context of an individual file).

But this does raise a good question -- what is our bar for exceptions to this rule of requiring autocorrect? 
  - For example, [switch-never-default](https://github.com/airbnb/swift#switch-never-default) and [extensions-for-protocol-conformance](https://github.com/airbnb/swift/issues/58) would both be incredibly challenging to autocorrect (they require complete knowledge of the codebase, whereas all existing linters operate on single files at a time).
  - [switch-never-default](https://github.com/airbnb/swift#switch-never-default) is still present in our style guide, despite not being autocorrectable, while [extensions-for-protocol-conformance](https://github.com/airbnb/swift/issues/58) was removed.

calda: Good point. I think in this case we should update the [namespace-using-enums](https://github.com/airbnb/swift#namespace-using-enums) rule to cover all cases, regardless of access control:

> Use caseless enums for organizing constants and functions into namespaces. Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.

calda: Fun fact that's tangentially related to autocorrecting [namespace-using-enums](https://github.com/airbnb/swift#namespace-using-enums):

The next version of SwiftFormat will have a rule that converts classes and structs to enums if they only contain static members (nicklockwood/SwiftFormat#737). 

This doesn't create _new_ namespaces, but it does help enforce the _Caseless enums work well as namespaces because they cannot be instantiated, which matches their intent_ part of that rule.

calda: Based on the discussion in https://github.com/airbnb/swift/pull/34#issuecomment-443930331, I'm leaning towards thinking that we wouldn't want to require autocorrect for a rule like this (even if autocorrect is theoretically possible). This rule seems more like a "pattern" than a way to "format" code.

calda: https://github.com/airbnb/swift/pull/106#discussion_r495209375

README.md
```diff
@@ -959,27 +959,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
-      print(privateValue)
-      print(MyClass.publicValue)
-    }
-  }
-  ```
-
-  </details>
-
-* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
+* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
```
bachand, line 962:
```suggestion
* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
```

If we go this direction (and I'm still not sure I'm 100% sold) I would recommend we make more clear in the first sentence that we are saying to prefer `enum`s to _global_ functions and constants.

bachand, line 962: I want to point out that now we are saying you can't have global `private` functions either.... 😓

README.md
```diff
@@ -959,7 +959,10 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
+* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Prefer using caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.**
+  * Feel free to nest namespaces where it adds clarity.
+  * Avoid creating non-namespaced global constants and functions.
+  * `private` globals are permitted, since they are scoped to a single file and do not pollute the global namespace. Consider placing private globals in an enum namespace to match the guidelines for other declaration types.
```
calda, line 965: @bachand I updated this based on our discussions offline

README.md
bachand: We discussed internally.

@fdiaz pointed us to [this comment](https://github.com/airbnb/swift/pull/34#issuecomment-443930331). We reminded ourselves of that, per our [guiding tenets](https://github.com/airbnb/swift#guiding-tenets), rules that _reformat code_ need to be able to be reformatted automatically (barring heavily justified exceptions). Per the examples in the linked comment, rules that _reformat_ code are those that _do not change the nature of the code_.

This is why [switch-never-default](https://github.com/airbnb/swift#switch-never-default) remains in our style guide while [extensions-for-protocol-conformance](https://github.com/airbnb/swift/issues/58) was removed. A very pure example of a rule that reformats code is one that adds `// MARK`s. [extensions-for-protocol-conformance](https://github.com/airbnb/swift/issues/58) does change code but doesn't change the meaning, so it falls at the extreme end of what we would consider reformatting. I can't think of how moving methods to an extension could cause a file to no longer compile.

bachand: Applying these learnings ^ to this example, we can see that this rule does not "reformat code". This is why it remains in the style guide and is **not** in violation of our guiding tenets.

README.md
```diff
@@ -959,27 +959,10 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
-      print(privateValue)
-      print(MyClass.publicValue)
-    }
-  }
-  ```
-
-  </details>
-
-* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.** Avoid creating non-namespaced global constants and functions. Feel free to nest namespaces where it adds clarity.
+* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Prefer using caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.**
```
bachand, line 962:
```suggestion
* <a id='namespace-using-enums'></a>(<a href='#namespace-using-enums'>link</a>) **Use caseless `enum`s for organizing `public` or `internal` constants and functions into namespaces.**
```

Let's keep this change focused on the private constants and leave this part the same?

README.md
```diff
@@ -959,27 +959,10 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  * Feel free to nest namespaces where it adds clarity.
+  * Avoid creating non-namespaced global constants and functions.
```
bachand, line 964: Can we keep the order the same (i.e. "Avoid creating non-namespaced..." before "Feel free to nest namespaces..."). I'm not sure it matters but I'd like this PR to stay focused on the private constants and make as few other changes as possible.

README.md
```diff
@@ -959,27 +959,10 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  * `private` globals are permitted, since they are scoped to a single file and do not pollute the global namespace. Consider placing private globals in an enum namespace to match the guidelines for other declaration types.
```
bachand, line 965:
```suggestion
  * `private` globals are permitted, since they are scoped to a single file and do not pollute the global namespace. Consider placing private globals in an `enum` namespace to match the guidelines for other declaration types.
```

... to match the earlier part of the rule

bachand (approved):
Based on revisions, I'm changing my 👎 to a 👍  and approving. I would prefer to wait for another stamp before we merge this. I am onboard with not _encouraging_ developers to add properties to the global namespace of a file, while still allowing it. As I mentioned internally, I think it's important to be very thoughtful if we are removing or heavily restricting a language feature. I think in some cases there is a lot of power in taking advantage of the the fact that global `private` members are only visible within a single file.

README.md
bachand: That may be true though I feel like it's hard to objectively quantify this.

bachand: I edited the first comment in this thread to indicate that we have updated our thinking, to save readers time.

bachand: We [revised](https://github.com/airbnb/swift/pull/106#discussion_r495278437) our thinking.

fdiaz (approved):

Issue #54, opened 2019-01-14
Each type in a file should be preceded by MARK


fdiaz:
This rule is missing autocorrect. Following [this PR](https://github.com/airbnb/swift/pull/35).

---

* <a id='marks-for-types'></a>(<a href='#marks-for-types'>link</a>) **Each type in a file should be preceded by `// MARK: - TypeName`.** [![SwiftLint: mark](https://img.shields.io/badge/SwiftLint-mark-008489.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#mark)

  <details>

  #### Why?
  The hyphen visually distinguishes types from sections within those types (described below).

  </details>

calda:
Added in #107

PR #107, opened 2020-10-06
Place a MARK comment before types and extensions

calda:
#### Summary

This PR proposes a new rule that each type and extension should be preceded by a `MARK` comment:
  * Types should be preceded by a `// MARK: - TypeName` comment.
  * Extensions that add a conformance should be preceded by a `// MARK: TypeName + ProtocolName` comment.
  * Extensions that follow the type being extended should omit that type's name and instead use `// MARK: ProtocolName`.
  * If there is only one type or extension in a file, the `MARK` comment can be omitted.
  * For extensions that do not add new conformances, consider adding a `// MARK:` with a descriptive comment.

  ```swift
  // MARK: - GalaxyView

  final class GalaxyView: UIView { }

  // MARK: ContentConfigurableView

  extension GalaxyView: ContentConfigurableView { }

  // MARK: Galaxy + SpaceThing

  extension Galaxy: SpaceThing { }
  ```

#### Reasoning

This rule is a combination of the lintable portions of two rules that used to be part of our style guide:
 - [Each type in a file should be preceded by MARK](https://github.com/airbnb/swift/issues/54)
 - [Each protocol conformance implementation should occur in dedicated type extension within the same file as the type](https://github.com/airbnb/swift/issues/58) (_"This extension should be marked with `// MARK: ProtocolName`"_)

It is now possible to autocorrect for this, using SwiftFormat's new `markTypes` rule ([nicklockwood/SwiftFormat#747](https://github.com/nicklockwood/SwiftFormat/pull/747)).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 * <a id='newline-at-eof'></a>(<a href='#newline-at-eof'>link</a>) **Files should end in a newline.** [![SwiftLint: trailing_newline](https://img.shields.io/badge/SwiftLint-trailing__newline-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-newline)
 
+* <a id='mark-types-and-extensions'></a>(<a href='#mark-types-and-extensions'>link</a>) **Each type and extension should be preceded by a `MARK` comment.** [![SwiftFormat: markTypes](https://img.shields.io/badge/SwiftFormat-markTypes-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#markTypes)
+  * Types should be preceded by a `// MARK: - TypeName` comment.
+  * Extensions that add a conformance should be preceded by a `// MARK: TypeName + ProtocolName` comment.
+  * Extensions that follow the type being extended should omit that type's name and instead use `// MARK: ProtocolName`.
+  * If there is only one type or extension in a file, the `MARK` comment can be omitted.
+  * For extensions that do not add new conformances, consider adding a `// MARK:` with a descriptive comment.
+
+  <details>
+
+  ```swift
+  // MARK: - GalaxyView
+
+  final class GalaxyView: UIView { }
+
+  // MARK: ContentConfigurableView
+
+  extension GalaxyView: ContentConfigurableView { }
+
+  // MARK: Galaxy + SpaceThing
+
+  extension Galaxy: SpaceThing { }
+  ```
+
+  </details>
+
 * <a id='marks-within-types'></a>(<a href='#marks-within-types'>link</a>) **Use `// MARK:` to separate the contents of a type definition into the sections listed below, in order.** All type definitions should be divided up in this consistent way, allowing a new reader of your code to easily jump to what they are interested in. [![SwiftFormat: organizeDeclarations](https://img.shields.io/badge/SwiftFormat-organizeDeclarations-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#organizeDeclarations)
   * `// MARK: Lifecycle` for `init` and `deinit` methods.
   * `// MARK: Open` for `open` properties and methods.


README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 * <a id='newline-at-eof'></a>(<a href='#newline-at-eof'>link</a>) **Files should end in a newline.** [![SwiftLint: trailing_newline](https://img.shields.io/badge/SwiftLint-trailing__newline-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-newline)
 
+* <a id='mark-types-and-extensions'></a>(<a href='#mark-types-and-extensions'>link</a>) **Each type and extension should be preceded by a `MARK` comment.** [![SwiftFormat: markTypes](https://img.shields.io/badge/SwiftFormat-markTypes-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#markTypes)
+  * Types should be preceded by a `// MARK: - TypeName` comment.
+  * Extensions that add a conformance should be preceded by a `// MARK: TypeName + ProtocolName` comment.
+  * Extensions that follow the type being extended should omit that type's name and instead use `// MARK: ProtocolName`.
+  * If there is only one type or extension in a file, the `MARK` comment can be omitted.
+  * For extensions that do not add new conformances, consider adding a `// MARK:` with a descriptive comment.
```
calda, line 1357: It's not really possible to autocorrect a descriptive `MARK` comment for extensions that don't add additional conformances, but it seems valuable to leave this in as a suggestion / best-practice.

bachand: How does the autocorrect rule handle this?

README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  final class GalaxyView: UIView { }
+
+  // MARK: ContentConfigurableView
+
+  extension GalaxyView: ContentConfigurableView { }
+
+  // MARK: Galaxy + SpaceThing
+
+  extension Galaxy: SpaceThing { }
```
bachand, line 1372: What happens for `extension Galaxy: SpaceThing, ContainsDarkMatter { }`?

calda: The [SwiftFormat rule](https://github.com/nicklockwood/SwiftFormat/pull/747/files#diff-68bc6d56aad81543185711914b2a1424R1675) uses a comma-separated list of the conformances:

```diff
+ // MARK: Galaxy + SpaceThing, DarkMatterContaining
+
 extension Galaxy: SpaceThing, DarkMatterContaining { }
```

README.md
calda: This can't be autocorrected, since the linter has no way of synthesizing a useful "descriptive comment". It seems useful to include optional guidance (_"consider adding"_) for what to do in this case.

An example of this in practice would be something like:

```swift
// MARK: Constraint helpers

extension UIView {
  func constrainToSuperview() { ... }
  func constrainToSuperviewSafeAreaInsets() { ... }
}
```

README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  * Extensions that follow the type being extended should omit that type's name and instead use `// MARK: ProtocolName`.
+  * If there is only one type or extension in a file, the `MARK` comment can be omitted.
+  * For extensions that do not add new conformances, consider adding a `// MARK:` with a descriptive comment.
+
+  <details>
+
+  ```swift
+  // MARK: - GalaxyView
+
+  final class GalaxyView: UIView { }
```
erichoracek, line 1364: nit: Include an ellipsis here and below to indicate that this declaration would have content that's been omitted for brevity
```suggestion
  final class GalaxyView: UIView { … }
```

README.md
erichoracek: If we're adding commas between protocols it feels like we should just have the mark contents match the Swift declaration since folks are familiar with that syntax, e.g. `Galaxy: SpaceThing, DarkMatterContaining`. Using a plus feels less precise, especially since we're only using it for protocol conformances and not other types of marks

README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  // MARK: ContentConfigurableView
+
+  extension GalaxyView: ContentConfigurableView { }
```
erichoracek, line 1368:
```suggestion
  extension GalaxyView: ContentConfigurableView { … }
```

README.md
erichoracek, line 1372:
```suggestion
  extension Galaxy: SpaceThing { … }
```

README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 * <a id='newline-at-eof'></a>(<a href='#newline-at-eof'>link</a>) **Files should end in a newline.** [![SwiftLint: trailing_newline](https://img.shields.io/badge/SwiftLint-trailing__newline-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-newline)
 
+* <a id='mark-types-and-extensions'></a>(<a href='#mark-types-and-extensions'>link</a>) **Each type and extension should be preceded by a `MARK` comment.** [![SwiftFormat: markTypes](https://img.shields.io/badge/SwiftFormat-markTypes-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#markTypes)
+  * Types should be preceded by a `// MARK: - TypeName` comment.
+  * Extensions that add a conformance should be preceded by a `// MARK: TypeName + ProtocolName` comment.
```
erichoracek, line 1354: Shouldn't this also have a dash since it's a standalone type and it should have its own section?
```suggestion
  * Extensions that add a conformance should be preceded by a `// MARK: - TypeName + ProtocolName` comment.
```

README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  ```swift
+  // MARK: - GalaxyView
+
+  final class GalaxyView: UIView { }
+
+  // MARK: ContentConfigurableView
+
+  extension GalaxyView: ContentConfigurableView { }
+
+  // MARK: Galaxy + SpaceThing
```
erichoracek, line 1370: Same as above—shouldn't this get a dash since this is the beginning of a new type grouping?
```suggestion
  // MARK: - Galaxy + SpaceThing
```

erichoracek (approved):
Overall LGTM! Just a few small suggestions/questions

README.md
calda: Good suggestion, that sounds reasonable. I can update the SwiftFormat rule to support this level of customization.

README.md
calda: Hmm good question. Do we want the rule to prefer using a dash _unless_ it's an extension following the type being extended?

The SwiftFormat rule is currently set up to support different comment templates for types (`// MARK: - %t` by default) and extensions (`// MARK: %t` by default). I would have to rethink how the SwiftFormat rule works to support this behavior, but am open to doing that if this is the direction we want to go in.

README.md
calda: although this does mean that extension marks would have two colons:

```swift
// MARK: Galaxy: SpaceThing, DarkMatterContainig
```

README.md
bachand: What does SwiftFormat do if I wrote something like this?

```swift

// MARK: Constraint helpers

extension UIView: RandomProtocol {
  func constrainToSuperview() { ... }
  func constrainToSuperviewSafeAreaInsets() { ... }
}
```

I'm curious how autocorrection works in that case.

README.md
bachand: I think we should prefer the `+`. Having two `:` characters feels odd. I think the `+` also is a convention people will understand. In the Objective-C days it was very common to use a `+` in file names when a file implemented a category (similar to a protocol), e.g. `MyObjCCode+Category.h`/`MyObjCCode+Category.m`.

README.md
bachand: I think it makes the most sense if we use `-` to indicate separation between types, so I agree with Eric's revision.

README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 * <a id='newline-at-eof'></a>(<a href='#newline-at-eof'>link</a>) **Files should end in a newline.** [![SwiftLint: trailing_newline](https://img.shields.io/badge/SwiftLint-trailing__newline-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-newline)
 
+* <a id='mark-types-and-extensions'></a>(<a href='#mark-types-and-extensions'>link</a>) **Each type and extension should be preceded by a `MARK` comment.** [![SwiftFormat: markTypes](https://img.shields.io/badge/SwiftFormat-markTypes-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#markTypes)
```
bachand, line 1352:
```suggestion
* <a id='mark-types-and-extensions'></a>(<a href='#mark-types-and-extensions'>link</a>) **Each type and extension which implements a conformance should be preceded by a `MARK` comment.** [![SwiftFormat: markTypes](https://img.shields.io/badge/SwiftFormat-markTypes-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#markTypes)
```

We're not requiring you to add marks when you have "anonymous" extensions (see the final bullet point). Should we be more precise?

README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  final class GalaxyView: UIView { … }
+
+  // MARK: ContentConfigurableView
+
+  extension GalaxyView: ContentConfigurableView { … }
+
+  // MARK: Galaxy + SpaceThing
+
+  extension Galaxy: SpaceThing { … }
```
bachand, line 1372: I think it could be nice to have an example of conforming to multiple protocols in a single extension.

README.md
calda: The SwiftFormat rule will replace any existing mark with the correct mark. In this case: 

```diff
- // MARK: Constraint helpers
+ // MARK: UIView + RandomProtocol

 extension UIView: RandomProtocol {
   func constrainToSuperview() { ... }
   func constrainToSuperviewSafeAreaInsets() { ... }
 }
```

README.md
calda: good point!

README.md
```diff
@@ -1349,6 +1349,31 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  * Extensions that follow the type being extended should omit that type's name and instead use `// MARK: ProtocolName`.
+  * If there is only one type or extension in a file, the `MARK` comment can be omitted.
+  * For extensions that do not add new conformances, consider adding a `// MARK:` with a descriptive comment.
+
+  <details>
+
+  ```swift
+  // MARK: - GalaxyView
+
+  final class GalaxyView: UIView { … }
```
bachand, line 1364: I think there is something nice to having code that you can copy into an editor and compile without changes, but I don't feel strongly against this.

bachand:
I'm generally supportive so I'm given this PR a 👍  though there are a few more things I'd like to work through before I'm ready to approve. Good work @calda !

README.md
calda: https://github.com/nicklockwood/SwiftFormat/pull/769 adds support for this behavior to the SwiftFormat `markTypes` rule.

README.md
erichoracek: I can see the argument for the plus—sounds good to me.

calda:
Thanks everyone for your reviews and support for this change! I'll merge this PR once we are able to adopt SwiftFormat 0.47+ (with the new `markTypes` rule)

README.md
```diff
@@ -1349,6 +1349,32 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 * <a id='newline-at-eof'></a>(<a href='#newline-at-eof'>link</a>) **Files should end in a newline.** [![SwiftLint: trailing_newline](https://img.shields.io/badge/SwiftLint-trailing__newline-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#trailing-newline)
 
+* <a id='mark-types-and-extensions'></a>(<a href='#mark-types-and-extensions'>link</a>) **Each type and extension which implements a conformance should be preceded by a `MARK` comment.** [![SwiftFormat: markTypes](https://img.shields.io/badge/SwiftFormat-markTypes-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#markTypes)
+  * Types should be preceded by a `// MARK: - TypeName` comment.
+  * Extensions that add a conformance should be preceded by a `// MARK: - TypeName + ProtocolName` comment.
+  * Extensions that follow the type being extended should omit that type's name and instead use `// MARK: ProtocolName`.
+  * If there is only one type or extension in a file, the `MARK` comment can be omitted.
+  * If the extension in question is empty (e.g. has no delcarations in its body), the `MARK` comment can be omitted.
```
calda, line 1357: I added this extra stipulation which matches the behavior of SwiftFormat's `markTypes` rule

There are several examples from our codebase where adding `MARK` comments to empty extensions creates a lot of unnecessary noise, like:

```swift
extension AppScope: APISessionProviding { }
extension AppScope: ExperimentsServiceProviding { }
extension AppScope: LocaleServiceProviding { }
// ...
```

bachand, line 1357:
```suggestion
  * If the extension in question is empty (e.g. has no declarations in its body), the `MARK` comment can be omitted.
```

PR #113, opened 2020-11-20
Update addTarget to the latest Swift style

zarochintsev:
#### Summary
I just updated `addTarget` to the latest Swift style.

#### Reasoning
Apple updated Swift style.


calda (approved):
Thanks!

PR #117, opened 2021-04-06
Adjust example code to the described rule.

kkiermasz:
#### Summary

* Fix "capitalize-acronyms" example
* Add missing parentheses in 'use-camel-case' example

#### Reasoning

The reasoning explanation is not necessary. It's just a bug/typo fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
 
   let urlValidator = URLValidator()
-  let isProfile = urlValidator.isProfileUrl(urlToTest, userID: idOfUser)
+  let isProfile = urlValidator.isProfileURL(urlToTest, userID: idOfUser)
   ```
 
   </details>
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
     public func executeRequest(
       _ request: URLRequest,
       onSuccess: @escaping (ModelType, Bool) -> Void,
-      onFailure: @escaping (Error) -> Void) -> URLSessionCancellable
+      onFailure: @escaping (Error) -> Void) -> URLSessionCancellable)
     {
       return _executeRequest(request, session, parser, onSuccess, onFailure)
     }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,7 +171,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
 
   let urlValidator = URLValidator()
-  let isProfile = urlValidator.isProfileUrl(urlToTest, userID: idOfUser)
+  let isProfile = urlValidator.isProfileURL(urlToTest, userID: idOfUser)
```
bachand, line 174: Wow this is a big oversight. Thanks!

fdiaz (approved):
Thank you for contributing!

README.md
```diff
@@ -108,15 +108,15 @@ _You can enable the following settings in Xcode by running [this script](resourc
     public func executeRequest(
       _ request: URLRequest,
       onSuccess: @escaping (ModelType, Bool) -> Void,
-      onFailure: @escaping (Error) -> Void) -> URLSessionCancellable
+      onFailure: @escaping (Error) -> Void) -> URLSessionCancellable)
```
bachand, line 111:
```suggestion
      onFailure: @escaping (Error) -> Void)
      -> URLSessionCancellable
```

So actually I think we want this. `URLSessionCancellable` is the return type of the `executeRequest(...)` function. BUT I'm super glad you flagged this, because here we aren't following [this](https://github.com/airbnb/swift#long-function-declaration) guideline, which may have led to the confusion.

README.md
```diff
@@ -108,15 +108,15 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
     {
       return _executeRequest(request, session, parser, onSuccess, onFailure)
     }
 
     private let _executeRequest: (
       URLRequest,
       @escaping (ModelType, Bool) -> Void,
-      @escaping (NSError) -> Void) -> URLSessionCancellable
+      @escaping (NSError) -> Void) -> URLSessionCancellable)
```
bachand, line 119:
```suggestion
      @escaping (NSError) -> Void)
      -> URLSessionCancellable
```

README.md
kkiermasz: Yes, you're right. It won't even compile now :P

bachand: Could you put up another PR? Sorry for the hassle.

kkiermasz: Sure!

PR #119, opened 2021-04-21
Add "No direct standard out" rule

erichoracek:
### Summary
**Prefer dedicated logging systems like [`os_log`](https://developer.apple.com/documentation/os/logging) or [`swift-log`](https://github.com/apple/swift-log) over writing directly to standard out using `print(…)`, `logDebug(…)`, or `dump(…)`.**

Additionally includes a custom SwiftLint regex rule to flag violations as warnings. This rule isn't possible to autocorrect in the general case as the correct logging method depends on the specific app.

### Reasoning
All log messages should flow into intermediary logging systems that can direct them to the correct destination(s) based on the app's environment or configuration. `print(…)`, `debugPrint(…)`, or `dump(…)` will write directly to standard out in all app configurations and can potentially leak personally identifiable information (PII).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1382,7 +1382,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   <details>
 
   #### Why?
-  All log messages should flow into intermediary logging systems that can direct them to the correct destination(s) based on the app's environment or configuration. `print(…)`, `debugPrint(…)`, or `dump(…)` will write directly to standard out in all app configurations and can potentially leak personally identifiable information (PII).
+  All log messages should flow into intermediate logging systems that can direct them to the correct destination(s) based on the app's environment or configuration. `print(…)`, `debugPrint(…)`, or `dump(…)` will write directly to standard out in all app configurations and can potentially leak personally identifiable information (PII).
 
   </details>
 
--- a/README.md
+++ b/README.md
@@ -1377,12 +1377,12 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='no-direct-standard-out-logs'></a>(<a href='#no-direct-standard-out-logs'>link</a>) **Prefer dedicated logging systems like [`os_log`](https://developer.apple.com/documentation/os/logging) or [`swift-log`](https://github.com/apple/swift-log) over writing directly to standard out using `print(…)`, `logDebug(…)`, or `dump(…)`.**
+* <a id='no-direct-standard-out-logs'></a>(<a href='#no-direct-standard-out-logs'>link</a>) **Prefer dedicated logging systems like [`os_log`](https://developer.apple.com/documentation/os/logging) or [`swift-log`](https://github.com/apple/swift-log) over writing directly to standard out using `print(…)`, `debugPrint(…)`, or `dump(…)`.**
 
   <details>
 
   #### Why?
-  All log messages should flow into intermediate logging systems that can direct them to the correct destination(s) based on the app's environment or configuration. `print(…)`, `debugPrint(…)`, or `dump(…)` will write directly to standard out in all app configurations and can potentially leak personally identifiable information (PII).
+  All log messages should flow into intermediate logging systems that can direct messages to the correct destination(s) and potentially filter messages based on the app's environment or configuration. `print(…)`, `debugPrint(…)`, or `dump(…)` will write all messages directly to standard out in all app configurations and can potentially leak personally identifiable information (PII).
 
   </details>
 
--- a/README.md
+++ b/README.md
@@ -1377,6 +1377,15 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='no_direct_standard_out'></a>(<a href='#no_direct_standard_out'>link</a>) **Prefer dedicated logging systems like [`os_log`](https://developer.apple.com/documentation/os/logging) or [`swift-log`](https://github.com/apple/swift-log) over writing logs directly to standard out using `print(…)`, `logDebug(…)`, or `dump(…)`.**
+
+  <details>
+
+  #### Why?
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1375,6 +1375,15 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='no_direct_standard_out'></a>(<a href='#no_direct_standard_out'>link</a>) **Prefer dedicated logging systems like [`os_log`](https://developer.apple.com/documentation/os/logging) or [`swift-log`](https://github.com/apple/swift-log) over writing directly to standard out using `print(…)`, `logDebug(…)`, or `dump(…)`.**
```
bachand, line 1378: For a Swift CLI tool, `print(...)` is appropriate (as I understand). Is the intention here to allow that use case, since writing to STDOUT is not "logging" in a CLI tool?

erichoracek: Good point—even in a Swift CLI I'd imagine it would be a best practice to support dynamic levels of output, e.g. a `--quiet` mode or a `--verbose` mode that includes extra levels of output. Just calling `print` directly feels like it would still be a code small. What do you think?

bachand: I agree that `--quiet` and `--verbose` modes are great additions to a robust CLI tool. Those modes would impact how much output you ultimately sent to STDOUT. When you have decided what output to send to STDOUT though, wouldn't you need to invoke `print(...)` to actually write the output to the STDOUT stream?  `print(...)` also allows you to send output to STDERR by specifying a text output stream.

Screen Shot 2021-04-22 at 5 58 12 PM

bachand, line 1378: This part of the `swift-log` README is interesting.
Screen Shot 2021-04-22 at 6 03 27 PM

erichoracek: > When you have decided what output to send to STDOUT though, wouldn't you need to invoke print(...) to actually write the output to the STDOUT stream

I think you'd probably `write` with a formatted log string on a standard out `TextOutputStream` directly, which is what swift-log does:

https://github.com/apple/swift-log/blob/3577a992d373d0e7f75d1fdc6e6e4570b60a1e1f/Sources/Logging/Logging.swift#L1124

However if you do end up wanting to call `print` directly in your logging system, it should be possible to silence the warning. Perhaps we could update the rule "details" to specify that only the logging system should write directly to standard out?

bachand: Interesting, I hadn't considered `write(...)`. This rule makes more sense to me now that I see `print(...)` is under "Debugging and Reflection".
image

bachand: I _think_ that I am OK with this rule as-is now that I have learned more about the available API. As the rule is written I think it would be reasonable for a CLI tool to `write(...)` core output to STDOUT outside of a logging system, since core output wouldn't be considered a log in my mind. I wouldn't consider the text outputted by `echo` to be a log per se, and it seems like it would be OK for `echo` to directly use `write(...)` without a logging system, in my interpretation of this rule. Does this align with your intention?

bachand, line 1378: Should we use hyphens between words in the slug like we do for other rules?

bachand: If so, maybe the rule slug could be `no-direct-log-output`?

resources/swiftlint.yml
```diff
@@ -38,3 +38,10 @@ custom_rules:
     regex: "@objcMembers"
     message: "Explicitly use @objc on each member you want to expose to Objective-C"
     severity: error
+  no_direct_standard_out:
+    name: "Writing directly to standard out disallowed"
+    regex: "(\\bprint|\\bdebugPrint|\\bdump|Swift\\.print|Swift\\.debugPrint|Swift\\.dump)\\s*\\("
```
bachand, line 43: TIL a space before the `(` is valid.
Screen Shot 2021-04-22 at 8 58 21 PM

README.md
erichoracek: Fixed in https://github.com/airbnb/swift/pull/119/commits/4a8d457185e58819bc34a2f6040cb488df7c4d22

README.md
```diff
@@ -1377,6 +1377,15 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='no-direct-standard-out-logs'></a>(<a href='#no-direct-standard-out-logs'>link</a>) **Prefer dedicated logging systems like [`os_log`](https://developer.apple.com/documentation/os/logging) or [`swift-log`](https://github.com/apple/swift-log) over writing directly to standard out using `print(…)`, `logDebug(…)`, or `dump(…)`.**
```
bachand, line 1380:
```suggestion
* <a id='no-direct-standard-out-logs'></a>(<a href='#no-direct-standard-out-logs'>link</a>) **Prefer dedicated logging systems like [`os_log`](https://developer.apple.com/documentation/os/logging) or [`swift-log`](https://github.com/apple/swift-log) over writing directly to standard out using `print(…)`, `debugPrint(…)`, or `dump(…)`.**
```
Do you mean this?

README.md
```diff
@@ -1377,6 +1377,15 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  <details>
+
+  #### Why?
+  All log messages should flow into intermediate logging systems that can direct them to the correct destination(s) based on the app's environment or configuration. `print(…)`, `debugPrint(…)`, or `dump(…)` will write directly to standard out in all app configurations and can potentially leak personally identifiable information (PII).
```
bachand, line 1385:
```suggestion
  All log messages should flow into intermediate logging systems that can direct messages to the correct destination(s) and potentially filter messages based on the app's environment or configuration. `print(…)`, `debugPrint(…)`, or `dump(…)` will write all messages directly to standard out in all app configurations and can potentially leak personally identifiable information (PII).
```

bachand: I wonder if we can capture the idea of log levels too.

bachand (approved):

README.md
erichoracek: Good catch!

fdiaz (approved):

fdiaz:
Thank you Eric! I'm merging this since overall folks seem to approve of it.

PR #120, opened 2021-04-22
Add "prefer immutable statics" rule

erichoracek:
## Summary
**Prefer immutable or computed static properties over mutable ones whenever possible.** Use stored `static let` properties or computed `static var` properties over stored `static var`s properties whenever possible, as stored `static var` properties are global mutable state.

  #### Why?
  Global mutable state increases complexity and makes it harder to reason about the behavior of applications. It should be avoided when possible.

  ```swift
  // WRONG
  enum Fonts {
    static var title = UIFont(…)
  }

  // RIGHT
  enum Fonts {
    static let title = UIFont(…)
  }
  ```

  ```swift
  // WRONG
  struct FeatureState {
    var count: Int

    static var initial = FeatureState(count: 0)
  }

  // RIGHT
  struct FeatureState {
    var count: Int

    static var initial: FeatureState {
      // Vend static properties that are cheap to compute
      FeatureState(count: 0)
    }
  }
  ```
## Correction

The reason that we can't autocorrect this rule is that there's multiple potential fixes for any given static variable, and the linter doesn't have the context to decide for you. Some global mutable state is necessary, some global properties should be computed, and other global properties should be stored.

As a follow-up, I'll potentially add a SwiftLint regex rule with a warning to flag global stored static vars as a warning via the following regex:
```regex
\bstatic\s+var\s+\S+(:\s+\S+)?\s+=
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1154,7 +1154,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='prefer-immutable-statics'></a>(<a href='#prefer-immutable-statics'>link</a>) **Prefer immutable or computed static variables over mutable ones whenever possible.** Use stored `static let` properties or computed `static var` properties over stored `static var`s properties whenever possible, as stored `static var` properties are global mutable state.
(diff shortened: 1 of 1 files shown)


bachand (approved):

fdiaz (approved):

fdiaz:
Thank you Eric! I'm merging this since overall folks seem to approve of it.

PR #126, opened 2021-05-12
Add rule to always inline let keyword when destructuring enum cases and tuples

calda:
#### Summary

This PR proposes a new rule to always inline the `let` keyword when destructuring enum cases and tupes

> When destructuring an enum case or a tuple, place the `let` keyword inline, adjacent to each individual property assignment

```swift
// WRONG
switch result {
case let .success(value):
  // ...
case let .error(errorCode, errorReason):
  // ...
}

// WRONG
guard let case .success(value) else { 
  return 
}

// RIGHT
switch result {
case .success(let value):
  // ...
case .error(let errorCode, let errorReason):
  // ...
}

// RIGHT
guard case .success(let value) else {
  return
}
```

#### Reasoning

 1. **Consistency**: We should prefer to either _always_ inline the `let` keyworkd or _never_ inline the `let` keyword. In Airbnb's Swift codebase, we [observed](https://github.com/airbnb/swift/pull/126#discussion_r631979244) that inline `let` is used far more often in practice (especially when destructuring enum cases with a single associated value).

 2. **Clarity**: Inlining the `let` keyword makes it more clear which identifiers are part of the conditional check and which identifiers are binding new variables, since the `let` keyword is always adjacent to the variable identifier. 

```swift
// `let` is adjacent to the variable identifier, so it is immediately obvious 
// at a glance that these identifiers represent new variable bindings
case .enumCaseWithSingleAssociatedValue(let string):
case .enumCaseWithMultipleAssociatedValues(let string, let int):

// The `let` keyword is quite far from the variable identifiers, 
// so its less obvious that they represent new variable bindings
case let .enumCaseWithSingleAssociatedValue(string):
case let .enumCaseWithMultipleAssociatedValues(string, int):
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,23 +541,52 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   <details>
 
-  ```swift
-  // WRONG
-  switch result {
-    case let .value(value):
+    ```swift
+    // WRONG
+    switch result {
+    case let .success(value):
       // ...
     case let .error(errorCode, errorReason):
       // ...
-  }
+    }
 
-  // RIGHT
-  switch result {
-    case .value(let value):
+    // WRONG
+    guard let case .success(value) else { 
+      return 
+    }
+
+    // RIGHT
+    switch result {
+    case .success(let value):
       // ...
     case .error(let errorCode, let errorReason):
       // ...
-  }
-  ```
+    }
+
+    // RIGHT
+    guard case .success(let value) else {
+      return
+    }
+    ```
+    
+    ### Why?
+    
+    1. **Consistency**: We should prefer to either _always_ inline the `let` keyworkd or _never_ inline the `let` keyword. In Airbnb's Swift codebase, we [observed](https://github.com/airbnb/swift/pull/126#discussion_r631979244) that inline `let` is used far more often in practice (especially when destructuring enum cases with a single associated value).
+    
+    2. **Clarity**: Inlining the `let` keyword makes it more clear which identifiers are part of the conditional check and which identifiers are binding new variables, since the `let` keyword is always adjacent to the variable identifier. 
+    
+    ```swift
+    // `let` is adjacent to the variable identifier, so it is immediately obvious 
+    // at a glance that these identifiers represent new variable bindings
+    case .enumCaseWithSingleAssociatedValue(let string):
+    case .enumCaseWithMultipleAssociatedValues(let string, let int):
+
+    // The `let` keyword is quite far from the variable identifiers, 
+    // so its less obvious that they represent new variable bindings
+    case let .enumCaseWithSingleAssociatedValue(string):
+    case let .enumCaseWithMultipleAssociatedValues(string, int):
+
+    ```
 
   </details>
 
--- a/README.md
+++ b/README.md
@@ -537,7 +537,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='destructure-inline-let'></a> (<a href='#destructure-inline-let'>link</a>) **When destructuring an enum case or a tuple, place the `let` keyword inline, adjacent to each individual property assignment.** [![SwiftFormat: hoistPatternLet](https://img.shields.io/badge/SwiftFormat-hoistPatternLet-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#hoistPatternLet)
+* <a id='inline-let-when-destructuring'></a> (<a href='#inline-let-when-destructuring'>link</a>) **When destructuring an enum case or a tuple, place the `let` keyword inline, adjacent to each individual property assignment.** [![SwiftFormat: hoistPatternLet](https://img.shields.io/badge/SwiftFormat-hoistPatternLet-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#hoistPatternLet)
 
   <details>
 
@@ -569,7 +569,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
     }
     ```
     
-    ### Why?
+    #### Why?
     
     1. **Consistency**: We should prefer to either _always_ inline the `let` keyworkd or _never_ inline the `let` keyword. In Airbnb's Swift codebase, we [observed](https://github.com/airbnb/swift/pull/126#discussion_r631979244) that inline `let` is used far more often in practice (especially when destructuring enum cases with a single associated value).
     
--- a/README.md
+++ b/README.md
@@ -537,6 +537,30 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -537,6 +537,30 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='destructure-inline-let'></a> (<a href='#* <a id='destructure-inline-let'></a> (<a href='#unnecessary-enum-arguments'>link</a>) **When destructuring an enum case or a tuple, place the `let` keyword inline, adjacent to each individual property assignment.** [![SwiftFormat: hoistPatternLet](https://img.shields.io/badge/SwiftFormat-hoistPatternLet-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#hoistPatternLet)
```
fdiaz, line 540: the format of the links at the beginning is off :P

calda: whoops, fixed!

README.md
```diff
@@ -537,6 +537,30 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='destructure-inline-let'></a> (<a href='#destructure-inline-let'>link</a>) **When destructuring an enum case or a tuple, place the `let` keyword inline, adjacent to each individual property assignment.** [![SwiftFormat: hoistPatternLet](https://img.shields.io/badge/SwiftFormat-hoistPatternLet-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#hoistPatternLet)
+
+  <details>
```
fdiaz, line 542: Let's add the reasoning to the rule itself too 🙏

calda: Do you think we should include the contrived third code example in the style guide document as well? I'm on the fence about that, because it feels a little too contrived to be worth including (but I wanted to include it in the PR for completeness, since it was brought up in the discussion on Slack)

bachand:
I am having trouble following 3) in the reasoning. Can you make it more clear to me?

README.md
```diff
@@ -537,6 +537,30 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='destructure-inline-let'></a> (<a href='#destructure-inline-let'>link</a>) **When destructuring an enum case or a tuple, place the `let` keyword inline, adjacent to each individual property assignment.** [![SwiftFormat: hoistPatternLet](https://img.shields.io/badge/SwiftFormat-hoistPatternLet-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#hoistPatternLet)
```
bachand, line 540: Putting here so we have a thread:

>  2. For consistency, we should prefer either _always_ inlining the `let` keywords, or _never_ inlining the keywords. Since enum payloads with a small number of associated values are much more common than enum payloads with a large number of associated values, we should prefer always inlining the `let` keywords.

For an enum with one associated value, the number of characters is the same in either style. For an enum with multiple associated values, the proposed style is more verbose. I don't follow how this proposed style prioritizes enum cases with fewer associated values.

bachand, line 540: Putting here so we have a thread:

> 1. This makes it more clear which identifiers are part of the conditional (if / case) check and which identifiers are binding new variables.

I am open to this idea though it would be easier to visualize. What do you think about adding a code example where we bind new variables to show how the proposed style is superior?

calda: In our codebase, when destructuring an enum case with a single associated value, inline let (`case .success(let string)`) is an order of magnitude more common than hoisted let (`case let .success(string)`).

| | Example | Regex | Count |
| --- | --- | ---- | --- |
| Inline let | `case .success(let string)` | `case \.[A-Za-z]+\(let [^,]*\):` | 6842 |
| Hoisted let | `case let .success(string)` | `case let \.[A-Za-z]+\([^,]*\):` | 901 |

Looking generally at the use of inline vs hoisted let overall (with any number of arguments), you can see that inline is overall 4x more common that hoisted let:

| | Example | Regex | Count |
| --- | --- | ---- | --- |
| Inline let | `case .success(` | `case \.[A-Za-z]+\(` | 9404 |
| Hoisted let | `case let .success(` | `case let \.[A-Za-z]+\(` | 2395 |

Because of this clear existing preference, if we are going to pick between the two options, I think we should choose to inline the `let` keywords.

calda: My main point here is that, with inline let, the `let` keyword is always adjacent to the variables's identifier (`let x`) rather than being off somewhere else in the line. Off the top of my head, this style of enum destructuring is the only case in Swift where the `let` doesn't appear adjacent to the variable.

I can add a code example, like:

```swift
// `let` is adjacent to the variable identifier
case .enumCaseWithSingleAssociatedValue(let string):
case .enumCaseWithMultipleAssociatedValues(let string, let int):

// The `let` keyword is quite far 
case let .enumCaseWithSingleAssociatedValue(string):
case let .enumCaseWithMultipleAssociatedValues(string, int):
```

bachand: This is very helpful.

bachand: IIRC inline let is the Xcode default (we should verify)? I think it would be nice to include some high-level reasoning in the rule (e.g. "Xcode autocomplete pushes you this direction and we've seen this in practice in a large code base").

bachand: I would also be interested to know how many of our case have multiple associated values where we could save characters by moving from the more common inline let to the hoisted let.

bachand: Would this rule apply to `guard` as well as `switch`? I didn't realize we could use the inline `let` syntax with `guard` but guess we can.
Screen Shot 2021-05-13 at 6 06 47 PM

calda: According to the numbers above, we have:
 - 7,743 cases with a single associated value (the sum of the two rows in the first table)
 - 4,056 cases with more than one associated value (the sum of the rows in the second table, subtracting the cases with a single associated value)

I would personally favor consistency (always placing the `let` keyword in the same place) over some sort of conditional logic based on the number of associated value cases.

calda: > Would this rule apply to guard as well as switch? 

Yes, this would apply to everywhere where you can pattern match and destructure enums (including `if`, `guard`, etc). I should add a non-`switch` example to the rule.

bachand: > I would personally favor consistency (always placing the `let` keyword in the same place) over some sort of conditional logic based on the number of associated value cases.

Yes I am in full agreement here. We should have a single consistent rule.

bachand: Sorry for not piecing that data together myself. It makes sense that people inline `let` less for cases with multiple associated values, since the hoisted `let` saves characters when there are multiple associated values.

bachand: I can get onboard with preferring inline let in call cases though I think it would be good to make the reasoning stronger in the rule.

calda:
> I am having trouble following 3) in the reasoning. Can you make it more clear to me?

I removed this example because I don't think it added much value

README.md
calda: Updated the rule and PR description.

calda: Updated the reasoning in the rule and PR description.

bachand: @calda I'm not seeing reasoning in the rule itself. Do you need to push a commit?
Screen Shot 2021-05-14 at 4 22 02 PM

calda: I made the changes locally but forgot to push them -- fixed!

README.md
```diff
@@ -537,6 +537,59 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+      // ...
+    }
+
+    // RIGHT
+    guard case .success(let value) else {
+      return
+    }
+    ```
+    
+    ### Why?
```
bachand, line 572:
```suggestion
    #### Why?
```
This seems to be what we're doing in other places.

bachand (approved):

PR #127, opened 2021-05-12
Add rule to prefer shorthand form for Optional, Array, and Dictionary

calda:
#### Summary

This PR proposes a rule to prefer the shorthand "sugared" form for common standard library types (`Optional`, `Array`, and `Dictionary`)

> **For standard library types with a canonical shorthand form (`Optional`, `Array`, `Dictionary`), prefer using the shorthand form over the full generic form.**

```swift
// WRONG
let optional: Optional<String> = nil
let array: Array<String> = []
let dictionary: Dictionary<String, Any> = [:]

// RIGHT
let optional: String? = nil
let array: [String] = []
let dictionary: [String: Any] = [:]
```

#### Reasoning

This aligns with what most folks already do in practice. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,6 +659,22 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+  * <a id='standard-library-type-shorthand'></a>(<a href='#standard-library-type-sugar'>link</a>) **For standard library types with a canonical shorthand form (`Optional`, `Array`, `Dictionary`), prefer using the shorthand form over the full generic form. [![SwiftFormat: typeSugar](https://img.shields.io/badge/SwiftFormat-typeSugar-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#typeSugar)
+
+  <details>
+
+  ```swift
+  // WRONG
+  let optional: Optional<String> = nil
+  let array: Array<String> = []
+  let dictionary: Dictionary<String, Any> = [:]
+
+  // RIGHT
+  let optional: String? = nil
+  let array: [String] = []
+  let dictionary: [String: Any] = [:]
+  ```
+
 ### Functions
 
 * <a id='omit-function-void-return'></a>(<a href='#omit-function-void-return'>link</a>) **Omit `Void` return types from function definitions.** [![SwiftLint: redundant_void_return](https://img.shields.io/badge/SwiftLint-redundant__void__return-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#redundant-void-return)
--- a/README.md
+++ b/README.md
@@ -659,7 +659,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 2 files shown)


resources/airbnb.swiftformat
```diff
@@ -22,4 +22,4 @@
 --swiftversion 5.1
 
 # rules
---rules anyObjectProtocol,redundantParens,redundantReturn,redundantSelf,sortedImports,strongifiedSelf,trailingCommas,trailingSpace,wrapArguments,wrapMultilineStatementBraces,indent,wrapAttributes,organizeDeclarations,markTypes,extensionAccessControl,duplicateImports,redundantType
+--rules anyObjectProtocol,redundantParens,redundantReturn,redundantSelf,sortedImports,strongifiedSelf,trailingCommas,trailingSpace,wrapArguments,wrapMultilineStatementBraces,indent,wrapAttributes,organizeDeclarations,markTypes,extensionAccessControl,duplicateImports,redundantType,typeSugar
```
bachand, line 25: Do you think it would be good to always specify the values to SwiftFormat (e.g. `--short-optionals always`) to avoid relying on SwiftFormat defaults which may change? I'm of mixed minds.

README.md
```diff
@@ -659,6 +659,22 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  ```swift
+  // WRONG
+  let optional: Optional<String> = nil
+  let array: Array<String> = []
+  let dictionary: Dictionary<String, Any> = [:]
+
+  // RIGHT
+  let optional: String? = nil
+  let array: [String] = []
+  let dictionary: [String: Any] = [:]
```
bachand, line 675: These examples violate [this](https://github.com/airbnb/swift#use-implicit-types) rule. Maybe we can use a different example?

bachand (commented):
This rule feels very aligned with the style guide overall :)

README.md
calda: Not sure what you mean. I don't think we have an explicit preference between

```
let optional: String? = nil
let array: [String] = []
let dictionary: [String: Any] = [:]
```

and 

```
let optional = String?.none
let array = [String]()
let dictionary = [String: Any]()
```

(if that's what you're getting at -- not sure if you have something else in mind)

fdiaz: I'm not sure I agree. These types cannot be inferred from the right hand side. I'd agree if the examples were using `let array: [String] = Array<String>()` but as is they need the explicit types.

README.md
```diff
@@ -659,6 +659,22 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+  * <a id='standard-library-type-shorthand'></a>(<a href='#standard-library-type-sugar'>link</a>) **For standard library types with a canonical shorthand form (`Optional`, `Array`, `Dictionary`), prefer using the shorthand form over the full generic form. [![SwiftFormat: typeSugar](https://img.shields.io/badge/SwiftFormat-typeSugar-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#typeSugar)
```
fdiaz, line 662: I think the format is off here too 😉

resources/airbnb.swiftformat
calda: Good question -- in general I think it makes sense to use SwiftFormat's default options (rather than preferring to always be explicit). There are a fair number of options, which apply to rules we have enabled, that we don't currently have an explicit choice for (and are relying on the default value).

README.md
bachand: Yes, I was mistaken.

README.md
calda: whoops, fixed

bachand (approved):

PR #130, opened 2021-06-10
Add rule that the opening brace following a single-line statement should be on the same line as the statement itself

calda:
#### Summary

This PR proposes a new rule that the opening brace following a single-line statement should be on the same line as the statement itself:

  ```swift
  // WRONG
  if !planet.isHabitable
  {
    planet.terraform()
  }

  class Planet
  {
    func terraform()
    {
      generateAtmosphere()
      generateOceans()
    }
  }

  // RIGHT
  if !planet.isHabitable {
    planet.terraform()
  }

  class Planet {
    func terraform() {
      generateAtmosphere()
      generateOceans()
    }
  }
  ```

#### Multi-line statements <a id='multi-line-statement-braces'></a>(<a href='#multi-line-statement-braces'>link</a>)

This PR also adds a prose rule for the multi-line case:

> The opening brace following a multi-line statement should wrap to a new line.

It felt like a good idea to add an explicit prose rule for this, to compliment the new rule for the single-statement case. This specific style is already mentioned in other rules, and we already lint for this automatically using the `wrapMultilineStatementBraces` SwiftFormat rule. 

  ```swift
  // WRONG
  if
    let star = planet.nearestStar(),
    planet.isInHabitableZone(of: star) {
    planet.terraform()
  }

  class Planet {
    func terraform(
      atmosphereOptions: AtmosphereOptions = .default,
      oceanOptions: OceanOptions = .default) {
      generateAtmosphere(atmosphereOptions)
      generateOceans(oceanOptions)
    }
  }

  // RIGHT
  if
    let star = planet.nearestStar(),
    planet.isInHabitableZone(of: star)
  {
    planet.terraform()
  }

  class Planet {
    func terraform(
      atmosphereOptions: AtmosphereOptions = .default,
      oceanOptions: OceanOptions = .default) 
    {
      generateAtmosphere(atmosphereOptions)
      generateOceans(oceanOptions)
    }
  }
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -839,7 +839,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='multi-line-expression-braces'></a>(<a href='#multi-line-expression-braces'>link</a>) The opening brace following a multi-line expression should wrap to a new line. [![SwiftFormat: wrapMultilineStatementBraces](https://img.shields.io/badge/SwiftFormat-wrapMultilineStatementBraces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapMultilineStatementBraces)
+* <a id='multi-line-expression-braces'></a>(<a href='#multi-line-expression-braces'>link</a>) The opening brace following a multi-line expression should wrap to a new line. [![SwiftFormat: braces](https://img.shields.io/badge/SwiftFormat-braces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#braces)
 
   <details>
 
@@ -901,7 +901,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='long-function-declaration'></a>(<a href='#long-function-declaration'>link</a>) **Separate [long](https://github.com/airbnb/swift#column-width) function declarations with line breaks before each argument label and before the return signature.** Put the open curly brace on the next line so the first executable line doesn't look like it's another parameter. [![SwiftFormat: wrapArguments](https://img.shields.io/badge/SwiftFormat-wrapArguments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapArguments) [![SwiftFormat: wrapMultilineStatementBraces](https://img.shields.io/badge/SwiftFormat-wrapMultilineStatementBraces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapMultilineStatementBraces)
+* <a id='long-function-declaration'></a>(<a href='#long-function-declaration'>link</a>) **Separate [long](https://github.com/airbnb/swift#column-width) function declarations with line breaks before each argument label and before the return signature.** Put the open curly brace on the next line so the first executable line doesn't look like it's another parameter. [![SwiftFormat: wrapArguments](https://img.shields.io/badge/SwiftFormat-wrapArguments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapArguments) [![SwiftFormat: braces](https://img.shields.io/badge/SwiftFormat-braces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#braces)
 
   <details>
 
--- a/README.md
+++ b/README.md
@@ -804,6 +804,83 @@ _You can enable the following settings in Xcode by running [this script](resourc
   let universe = Universe()
   ```
 
+* <a id='single-line-expression-braces'></a>(<a href='#single-line-expression-braces'>link</a>) The opening brace following a single-line expression should be on the same line as the rest of the statement. [![SwiftFormat: braces](https://img.shields.io/badge/SwiftFormat-braces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#braces)
+
+  <details>
+
+  ```swift
+  // WRONG
+  if !planet.isHabitable
+  {
+    planet.terraform()
+  }
+
+  class Planet
+  {
+    func terraform()
+    {
+      generateAtmosphere()
+      generateOceans()
+    }
+  }
+
+  // RIGHT
+  if !planet.isHabitable {
+    planet.terraform()
+  }
+
+  class Planet {
+    func terraform() {
+      generateAtmosphere()
+      generateOceans()
+    }
+  }
+  ```
+
+  </details>
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -728,6 +728,41 @@ _You can enable the following settings in Xcode by running [this script](resourc
   let dictionary: [String: Any] = [:]
   ```
 
+* <a id='single-line-expression-braces'></a>(<a href='#single-line-expression-braces'>link</a>) The opening brace following a single-line expression should not wrap to a newline.[![SwiftFormat: braces](https://img.shields.io/badge/SwiftFormat-braces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#braces)
```
bachand, line 731: Do you have any thoughts on what you would propose for multi-line expressions? I think starting here is great. I'm wondering what direction we may go to understand if it may inform this rule.

bachand (approved):

README.md
calda: Style for multi-line expressions is somewhat covered by other existing rules (and we have the SwiftFormat `wrapMultilineStatementBraces` rule enabled)
- "Put the open curly brace on the next line so the first executable line doesn't look like it's another parameter.": https://github.com/airbnb/swift#long-function-declaration
- Not explicitly mentioned in the prose, but included in examples: https://github.com/airbnb/swift#multi-line-conditions

It couldn't hurt to explicitly mention the multiline case in this rule, if that seems like it would be more clear.

bachand: > It couldn't hurt to explicitly mention the multiline case in this rule, if that seems like it would be more clear.

I like the idea of being explicit about braces for all types of expressions in this rule, though it seems like SwiftFormat will only autocorrect single-line expressions?

README.md
```diff
@@ -728,6 +728,83 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    func terraform() {
+      generateAtmosphere()
+      generateOceans()
+    }
+  }
+  ```
+
+  </details>
+
+* <a id='multi-line-statement-braces'></a>(<a href='#multi-line-statement-braces'>link</a>) The opening brace following a multi-line statement should wrap to a new line. [![SwiftFormat: wrapMultilineStatementBraces](https://img.shields.io/badge/SwiftFormat-wrapMultilineStatementBraces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapMultilineStatementBraces)
```
calda, line 766: The style in this rule is already described in other places throughout the style guide (and we already lint for this automatically, since we have the `wrapMultilineStatementBraces` rule enabled). 

It felt like a good idea to add an explicit prose rule for this, to compliment the new rule for the single-statement case.

README.md
calda: I added a prose rule for the multi-line case in d2860b9

README.md
```diff
@@ -728,6 +728,83 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  // RIGHT
+  if !planet.isHabitable {
+    planet.terraform()
+  }
+
+  class Planet {
+    func terraform() {
+      generateAtmosphere()
+      generateOceans()
```
fdiaz, line 759:
```suggestion
      generateOceans()
      pollutePlanet()
      findNewPlanet()
```

fdiaz (approved):
LGTM!

README.md
calda: seems wasteful to terraform, pollute, and then discard a planet all in a single run loop cycle -- maybe we'd want to `await pollutePlanet()` 🤪

README.md
```diff
@@ -728,6 +728,83 @@ _You can enable the following settings in Xcode by running [this script](resourc
   let dictionary: [String: Any] = [:]
   ```
 
+* <a id='single-line-statement-braces'></a>(<a href='#single-line-statement-braces'>link</a>) The opening brace following a single-line statement should be on the same line as the rest of the statement. [![SwiftFormat: braces](https://img.shields.io/badge/SwiftFormat-braces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#braces)
```
bachand, line 731: Asking to learn: is there an official definition of a "statement" in Swift?

README.md
bachand: I think that it is reasonable to repeat this a bit in `long-function-declaration`.

bachand (approved):
Nice work @calda

README.md
calda: [Statements](https://docs.swift.org/swift-book/ReferenceManual/Statements.html#) and [expressions](https://docs.swift.org/swift-book/ReferenceManual/Expressions.html) have different meanings in the Swift Grammar. I think an "expression" specifically refers to something that can be the left-hand-side of an assignment operator.

bachand: From looking at these resources (thanks!) `expression` feels more appropriate in this case. As an example, it seems like an "if statement" includes all of the code starting with `if` and ending at the final `}`. If that's true, most if statements would not be a single line unless you had something like `if myBoolean { print("hello" }`.

Screen Shot 2021-06-15 at 7 09 40 PM

calda: Good point -- I think you're right on this. Updated to refer to these as "expressions".

README.md
calda: In `long-function-declaration`, I think this point is covered by:

> Put the open curly brace on the next line so the first executable line doesn't look like it's another parameter.

README.md
```diff
@@ -728,6 +728,83 @@ _You can enable the following settings in Xcode by running [this script](resourc
   let dictionary: [String: Any] = [:]
   ```
 
+* <a id='single-line-expression-braces'></a>(<a href='#single-line-expression-braces'>link</a>) The opening brace following a single-line expression should be on the same line as the rest of the statement. [![SwiftFormat: braces](https://img.shields.io/badge/SwiftFormat-braces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#braces)
```
bachand, line 731:
```suggestion
* <a id='single-line-expression-braces'></a>(<a href='#single-line-expression-braces'>link</a>) The opening brace following a single-line expression should be on the same line as the rest of the expression. [![SwiftFormat: braces](https://img.shields.io/badge/SwiftFormat-braces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#braces)
```
?

bachand (approved):

README.md
calda: Whoops, thanks!

calda:
I haven't merged this yet because I'm waiting for changes in [nicklockwood/SwiftFormat#955](https://github.com/nicklockwood/SwiftFormat/pull/955) to be released (likely in the next major SwiftFormat version).

resources/airbnb.swiftformat
```diff
@@ -52,7 +52,7 @@
 --rules wrap
 --rules wrapArguments
 --rules wrapAttributes
---rules wrapMultilineStatementBraces
```
calda, line 55: The `wrapMultilineStatementBraces` rule was removed in nicklockwood/SwiftFormat#1071 in favor of an option on the `braces` rule (which is enabled by default). So this change preserves the existing behavior for multi-line statements/expressions, and now also applies to single-line statements/expressions.

PR #132, opened 2021-08-25
Enable SwiftFormat wrap rule with an enforced max width of 130 characters

calda:
#### Summary

This PR enables the SwiftFormat [`wrap`](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrap) rule in `airbnb.swiftformat`. We would begin enforcing a maximum line width of _130 characters_, and automatically wrap code to sit within the max line width. We would continue _recommending_ (but not enforcing) a maximum line width of 100 characters.

#### Reasoning

We should begin enforcing our maximum column width rule with a linter (_"We strive to make every rule lintable"_).

SwiftFormat's `wrap` rule is not perfect, but it follows our style guide / existing informal style preferences a large majority of the time (in general, it is more successful at wrapping longer lines). 100 characters seems [too strict](https://github.com/airbnb/swift/pull/132#issuecomment-905693965) to enforce with the SwiftFormat `wrap` rule at this time, given its current behavior. Enforcing a "more generous" maximum width of 130 characters results in a fewer cases of "weird" wrapping, and is better than the status quo (where we make a recommendation, but enforce nothing).

### DIFF
--- a/resources/xcode_settings.bash
+++ b/resources/xcode_settings.bash
@@ -8,5 +8,4 @@ defaults write com.apple.dt.Xcode DVTTextEditorTrimWhitespaceOnlyLines -bool YES
 defaults write com.apple.dt.Xcode DVTTextIndentTabWidth -int 2
 defaults write com.apple.dt.Xcode DVTTextIndentWidth -int 2
 
-defaults write com.apple.dt.Xcode DVTTextShowPageGuide -bool true
-defaults write com.apple.dt.Xcode DVTTextPageGuideLocation -int 130
+defaults write com.apple.dt.Xcode DVTTextPageGuideLocation -int 100
--- a/resources/xcode_settings.bash
+++ b/resources/xcode_settings.bash
@@ -8,4 +8,5 @@ defaults write com.apple.dt.Xcode DVTTextEditorTrimWhitespaceOnlyLines -bool YES
 defaults write com.apple.dt.Xcode DVTTextIndentTabWidth -int 2
 defaults write com.apple.dt.Xcode DVTTextIndentWidth -int 2
 
-defaults write com.apple.dt.Xcode DVTTextPageGuideLocation -int 100
+defaults write com.apple.dt.Xcode DVTTextShowPageGuide -bool true
+defaults write com.apple.dt.Xcode DVTTextPageGuideLocation -int 130
--- a/README.md
+++ b/README.md
@@ -39,12 +39,12 @@ Note that brevity is not a primary goal. Code should be made more concise only i
 
 _You can enable the following settings in Xcode by running [this script](resources/xcode_settings.bash), e.g. as part of a "Run Script" build phase._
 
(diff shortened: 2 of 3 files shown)


resources/xcode_settings.bash
```diff
@@ -8,4 +8,5 @@ defaults write com.apple.dt.Xcode DVTTextEditorTrimWhitespaceOnlyLines -bool YES
 defaults write com.apple.dt.Xcode DVTTextIndentTabWidth -int 2
 defaults write com.apple.dt.Xcode DVTTextIndentWidth -int 2
 
-defaults write com.apple.dt.Xcode DVTTextPageGuideLocation -int 100
+defaults write com.apple.dt.Xcode DVTTextShowPageGuide -bool true
```
calda, line 11: Since we would be enforcing this maximum with a linter, I think it makes sense to show the page guide by default.

erichoracek:
> Since we have never enforced a maximum column width of 100 characters, it has never really been followed in practice. In my opinion, 100 chars is too strict to enforce automatically with a linter. Within our codebase, it appears that folks have generally followed a maximum character width of ~120-130. Adopting a "more generous" maximum width of 130 characters also results in fewer cases of "weird" auto-wrapping.

Personally I find that 120-130 characters is simply too long to comfortably fit two editor panes side-by-side on a laptop display, and furthermore I find that lines over 100 characters in length are much harder to read, which is generally in line with [typography recommendations](https://practicaltypography.com/line-length.html):

> Shorter lines are more comfortable to read than longer lines. As line length increases, your eye has to travel farther from the end of one line to the beginning of the next, making it harder to track your progress vertically. Aim for an average line length of 45–90 characters, including spaces. You can check line length using word count.

--- 

> Adopting a "more generous" maximum width of 130 characters also results in fewer cases of "weird" auto-wrapping.

I posit that once we start linting for line length folks will see the lint failures and clean up cases where the linter wrapped code in a "weird" way that made things confusing. It might be helpful to discuss a few real-world examples where the 100 character max resulted in weird or less readable results so we can discuss whether there's ways to make them better without adjusting the max line length.

calda:
I think a max width of 100 characters is a good recommendation, but I'm not sure it's suitable as a strictly-enforced maximum.

When evaluating style guide changes, I think it's important to consider how folks are currently writing code in practice. While we have historically recommended a max width of 100 characters, this hasn't been followed very broadly.

One reasonable proxy metric for current practices is the total diff count of a commit that reformats our codebase to follow a given max character width:

| Max Width | Diff Count | Files Affected |
| ----- | ---- | ----- |
| 100 | +165,000 / -64,000 | 10,000 files |
| 120 | +66,500 / -26,000 | 6,000 files |
| 130 | +45,000 / -18,000 | 4,600 files |
| 160 | +17,675 / - 8,502 | 2,369 files |

Enforcing a max-character width of 100 results in a significantly more changes to our codebase, which suggests to me that it is too strict. I think 130 is a reasonable choice, out of these options, since it reformats lines that are "obviously too long" without being too invasive to existing practices and preferences.

bachand (approved):

PR #135, opened 2021-09-17
Add avoid-redundant-closures rule

calda:
#### Summary

This PR adds a new `avoid-redundant-closures`, linted using the new SwiftFormat `redundantClosures` rule (added in nicklockwood/SwiftFormat#998)

> **Avoid closures that are always called immediately**. Instead, prefer inlining the value directly.

```swift
// WRONG
lazy var universe: Universe = { 
  Universe() 
}()

lazy var stars = {
  universe.generateStars(
    at: location,
    count: 5,
    color: starColor,
    withAverageDistance: 4)
}()

// RIGHT
lazy var universe = Universe() 

lazy var stars = universe.generateStars(
  at: location,
  count: 5,
  color: starColor,
  withAverageDistance: 4)
```

#### Reasoning

I see this pattern pretty frequently (thousands of times in our codebase), but it isn't necessary in most cases. I think this is just the first pattern folks reach for when writing a `lazy var`.


README.md
```diff
@@ -1497,6 +1497,36 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='avoid-redundant-closures'></a>(<a href='#avoid-redundant-closures'>link</a>) **Avoid closures that are always called immediately**. Instead, prefer inlining the value directly. [![SwiftFormat: redundantClosure](https://img.shields.io/badge/SwiftFormat-redundantClosure-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantClosure)
```
erichoracek, line 1500: The way that this is phrased I'm wondering if it would also apply to something like this:

```swift
func doSomething() {
  let result1 = {
    // inline work, using some local variables
  }()

  let result2 = {
    // similar inline work, using the same variable names
  }()
}
```

As-is I think if were to autocorrect this we'd potentially get compilation errors if it were applied to cases where the closures included variables with the same name as removing the enclosing closures would cause a compilation error. 

Perhaps we should scope this down to something like "Avoid single-expression closures that are always called immediately"?

erichoracek (commented):
Nice addition! Just one suggestion

README.md
calda: The SwiftFormat rule handles this correctly (it only reformats single-expression closures). It makes sense to be more explicit about this in the prose. Updated!

erichoracek: Got it—I think it would help to be clear about it especially since we don't have any counterexamples in the details 👍

erichoracek (approved):

PR #146, opened 2021-12-07
Use SwiftFormat redundantPattern rule instead of SwiftLint empty_enum_arguments rule

calda:
#### Summary

The PR adopts the SwiftFormat `redundantPattern` rule, instead of the SwiftLint `empty_enum_arguments` rule. 

#### Reasoning

Both of these rule implementations cover our existing [unnecessary-enum-arguments](https://github.com/airbnb/swift#unnecessary-enum-arguments) rule, but I noticed that the SwiftLint implementation misses some examples that the SwiftFormat implementation autocorrects properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -515,7 +515,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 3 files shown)


erichoracek (approved):

bachand (commented):
Thanks!

PR #147, opened 2021-12-07
Add rule that braces should be surrounded by whitespace (either a space or newline)

calda:
#### Summary

This PR adds a new rule that braces should be surrounded by whitespace (either a space or newline):

```swift
// WRONG
struct Planet{
  …
}

// WRONG
if condition{
  …
}else{
  …
}

// RIGHT
struct Planet {
  …
}

// RIGHT
if condition {
  …
} else {
  …
}
```

<a id='closure-brace-spacing'></a>(<a href='#closure-brace-spacing'>link</a>) This PR also updates the existing rule for how to handle closure braces (which is sightly more nuanced):

> **Closures should have a space or newline inside each brace.** Trailing closures should have a space or newline outside each brace.

```swift
// WRONG
let evenSquares = numbers.filter{$0.isMultiple(of: 2)}.map{  $0 * $0  }

// RIGHT
let evenSquares = numbers.filter { $0.isMultiple(of: 2) }.map { $0 * $0 }

// WRONG
let evenSquares = numbers.filter( { $0.isMultiple(of: 2) } ).map( { $0 * $0 } )

// RIGHT
let evenSquares = numbers.filter({ $0.isMultiple(of: 2) }).map({ $0 * $0 })

// WRONG
let evenSquares = numbers
  .filter{ 
    $0.isMultiple(of: 2) 
  }
  .map{ 
    $0 * $0 
  }

// RIGHT
let evenSquares = numbers
  .filter {
    $0.isMultiple(of: 2) 
  }
  .map {
    $0 * $0 
  }
```

These behaviors are implemented in the existing [`spaceInsideBraces`](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#spaceInsideBraces) and [`spaceAroundBraces`](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#spaceAroundBraces) SwiftFormat rules.

#### Reasoning

This matches the style that most folks are already following. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,13 +503,13 @@ _You can enable the following settings in Xcode by running [this script](resourc
   // WRONG
   if (userCount > 0) { ... }
   switch (someValue) { ... }
-  let evens = userCounts.filter { (number) in number % 2 == 0 }
+  let evens = userCounts.filter { (number) in number.isMultiple(of: 2) }
   let squares = userCounts.map() { $0 * $0 }
 
   // RIGHT
   if userCount > 0 { ... }
   switch someValue { ... }
-  let evens = userCounts.filter { number in number % 2 == 0 }
+  let evens = userCounts.filter { number in number.isMultiple(of: 2) }
   let squares = userCounts.map { $0 * $0 }
   ```
 
@@ -792,6 +792,8 @@ _You can enable the following settings in Xcode by running [this script](resourc
   let dictionary: [String: Any] = [:]
   ```
 
+  </details>
+
 * <a id='omit-explicit-init'></a>(<a href='#omit-explicit-init'>link</a>) **Omit explicit `.init` when not reqired.** [![SwiftFormat: redundantInit](https://img.shields.io/badge/SwiftFormat-redundantInit-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantInit)
 
   <details>
@@ -803,6 +805,8 @@ _You can enable the following settings in Xcode by running [this script](resourc
   // RIGHT
   let universe = Universe()
   ```
+
+  </details>
 
 * <a id='single-line-expression-braces'></a>(<a href='#single-line-expression-braces'>link</a>) The opening brace following a single-line expression should be on the same line as the rest of the statement. [![SwiftFormat: braces](https://img.shields.io/badge/SwiftFormat-braces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#braces)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1042,16 +1072,40 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
-* <a id='closure-brace-spacing'></a>(<a href='#closure-brace-spacing'>link</a>) **Single-line closures should have a space inside each brace.** [![SwiftLint: closure_spacing](https://img.shields.io/badge/SwiftLint-closure__spacing-007A87.svg)](https://github.com/realm/SwiftLint/blob/master/Rules.md#closure-spacing)
+* <a id='closure-brace-spacing'></a>(<a href='#closure-brace-spacing'>link</a>) **Closures should have a space or newline inside each brace.** Trailing closures should have a space or newline outside each brace. [![SwiftFormat: spaceInsideBraces](https://img.shields.io/badge/SwiftFormat-spaceInsideBraces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#spaceInsideBraces) [![SwiftFormat: spaceAroundBraces](https://img.shields.io/badge/SwiftFormat-spaceAroundBraces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#spaceAroundBraces)
```
bachand, line 1075: This is an issue with the existing rule but... do you think we should clarify that we want a _single_ space or newline. If so let's update both rules.

README.md
```diff
@@ -1042,16 +1072,40 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  // RIGHT
+  let evenSquares = numbers.filter { $0.isMultiple(of: 2) }.map { $0 * $0 }
+
+  // WRONG
+  let evenSquares = numbers.filter( { $0.isMultiple(of: 2) } ).map( { $0 * $0 } )
 
   // RIGHT
-  let evenSquares = numbers.filter { $0 % 2 == 0 }.map { $0 * $0 }
+  let evenSquares = numbers.filter({ $0.isMultiple(of: 2) }).map({ $0 * $0 })
```
bachand, line 1090: There seems to be an implicit rule here to drop parentheses and prefer the trailing closure syntax.

README.md
calda: Good point!

README.md
calda: It would definitely be nice to have a rule for when to prefer trailing closures or not. Although there are some complications that may make this impossible to autocorrect with current tooling:
 - imho it's typically desirable to _not_ use trailing closures when the closure argument has a semantically meaningful argument label. As one example, I prefer `contains(where: { $0.isTrue })` over `contains { $0.isTrue }`.
 - It's trivial to autocorrect to change code to be a trailing closure, but not the other way around (since this requires knowledge of the argument label that isn't present).

calda: I've considered proposing a rule to cover this topic, even without autocomplete support. Any thoughts?

bachand: I personally prefer the trailing closure syntax almost universally, though I see where you're coming from regarding there being additional clarity in including the argument in some cases. I wonder if we could scope down the problem. For example, if a function ends in a single closure with the argument label `completion` or `completionHandler` we could convert the callsite to use the trailing closure syntax. The downside of a rule like this is that it becomes a bit more arbitrary than some of our other rules IMHO. Another approach could be to convert functions that have more than 2 arguments where the last argument is a closure (with any label) to use the trailing closure syntax. This second version feels less arbitrary and it may be a reasonable enough heuristic for filtering out cases where the presence of the argument label meaningfully increases clarity at the callsite.

bachand: I also take back what I said in my previous comment https://github.com/airbnb/swift/pull/147/files#r764530898. I now see that we're not offering an opinion in these examples of whether to prefer the trailing closure syntax, since we stay consistent (using it or not) in the same RIGHT/WRONG pair. I was reviewing on mobile before and must have missed that.

erichoracek (approved):

PR #149, opened 2022-01-19
Adopt redundantFileprivate SwiftFormat rule

calda:
#### Summary

This PR enables the [`redundantFileprivate`](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantFileprivate) SwiftFormat rule.

#### Reasoning

We already have a prose rule for this in the style guide ([_"Access control should be at the strictest level possible: prefer `private` to `fileprivate`"_](https://github.com/airbnb/swift#limit-access-control)), so this just adds autocorrect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1382,7 +1382,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   ```swift
   // WRONG
-  struct Spaceship {
+  public struct Spaceship {
     // WRONG: `engine` is used in `extension Spaceship` below,
     // but extensions in the same file can access `private` members.
     fileprivate let engine: AntimatterEngine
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1376,7 +1376,60 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    // WRONG: `hull` is not used by any other type, so `fileprivate` is unnecessary. 
+    fileprivate let hull: Hull
+
+    // RIGHT: `navigation` is used in `extension Pilot` below,
+    // so `fileprivate` is necessary here.
+    fileprivate let navigation: SpecialRelativityNavigationService
+  }
+
+  extension Spaceship {
+    func blastOff() {
```
bachand, line 1399: It may be valuable to clarify that this is used elsewhere in the module, which is why it needs to be internal.

bachand (approved):

README.md
calda: I made these declarations `public`, which I think clarifies what you were getting at without having to explicitly add prose in a comment

PR #153, opened 2022-01-20
Use SwiftFormat redundantRawValues instead of SwiftLint redundant_string_enum_value

calda:
This PR removes the SwiftLint [`redundant_string_enum_value`](https://realm.github.io/SwiftLint/redundant_string_enum_value.html) rule (which doesn't support autocorrect) in favor of the SwiftFormat [`redundantRawValues`](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantrawvalues) rule (which does support autocorrect).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1486,7 +1486,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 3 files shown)


erichoracek (approved):

fdiaz (approved):

bachand (commented):
Thanks!

PR #157, opened 2022-02-09
Add rule that comment blocks should use single-line comments (// and ///)

calda:
#### Summary

This PR proposes a new rule that comment blocks should use single-line comments (`//` and `///`) rather than multi-line comments (`/* ... */` and `/** ... */`). This is implemented by SwiftFormat's [`blockComments`](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments) rule.

```swift
// WRONG

/**
* A planet that exists somewhere in the universe.
*
* Planets have many properties. For example, the best planets
* have atmospheres and bodies of water to support life.
*/
class Planet {
  /**
    Terraforms the planet, by adding an atmosphere and ocean
    that is hospitable for life.
    - Precondition: the planet does not already have an atmosphere.
      Adding another atmosphere to a planet that already has one is a very bad idea.
  */
  func terraform() {
    /* 
    Generate the atmosphere first, before generating the ocean.
    Otherwise, the water will just boil off immediately.
    */
    generateAtmosphere()

    /* Now that we have an atmosphere, it's safe to generate the ocean */
    generateOceans()
  }
}

// RIGHT

///
/// Planets have many properties. For example, the best planets
/// have atmospheres and bodies of water to support life.
class Planet {
  /// Terraforms the planet, by adding an atmosphere and ocean
  /// that is hospitable for life.
  ///  - Precondition: the planet does not already have an atmosphere.
  ///    Adding another atmosphere to a planet that already has one is a very bad idea.
  func terraform() {
    // Generate the atmosphere first, before generating the ocean.
    // Otherwise, the water will just boil off immediately.
    generateAtmosphere()

    // Now that we have an atmosphere, it's safe to generate the ocean
    generateOceans()
  }
}
```

#### Reasoning

 - In general, when there are multiple equivalent ways to express something, we introduce a style guide rule that prefers one over the other. These two forms of writing comments are equivalent, so it makes sense for us to have a rule that explicitly prefers one over the other.

 - In our codebase, single-line comments are significantly more common than multi-line comments (I don't have any firm data for this but could figure out a way to analyze this quantitatively if necessary). 

 - Single-line comments are the default in Xcode:
   - when using Xcode's "Comment selection" command (`⌘/`), it inserts single-line comments (`//`)
   - when using Xcode's "Generate documentation" command (`⌘-option-/`), it inserts single-line doc comments (`///`)

      

   - Xcode also automatically inserts single-line comments when adding a new line to a comment block. (I think one potential reason to prefer mutli-line comments is that it reduces the need to do comment bookkeeping when adding newlines, but this isn't a problem in modern versions of Xcode).
  
      

 - There are many different ways to structure / indent comments using `/* ... */`, but only one way to structure / indent comments using `//` (so these are easier to use and better for consistency):
      ```swift
      // this is a single-line comment
      ```

      ```swift
      /* this is a multi-line comment */

      /* 
      this is a multi-line comment 
      */

      /* 
       * this is a multi-line comment 
       */

      /* 
        this is a multi-line comment 
       */
      ``` 

### DIFF
--- a/README.md
+++ b/README.md
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='comments'></a>(<a href='#comments'>link</a>) **Comment blocks should use single-line comments (`//` and `///`)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
+
+  <details>
+
+  ```swift
+  // WRONG
+
+  /**
+  * A planet that exists somnewhere in the universe.
+  *
+  * Planets have many properties. For example, the best planets
+  * have atmospheres and bodies of water to support life.
+  */
+  class Planet {
+    /**
+      Terraforms the planet, by adding an atmosphere and ocean
+      that is hospitable for life.
+      - Precondition: the planet does not already have an atmosphere.
+        Adding another atmosphere to a planet that already has one is a very bad idea.
+    */
+    func terraform() {
+      /* 
+      Generate the atmosphere first, before generating the ocean.
+      Otherwise, the water will just boil off immediately.
+      */
+      generateAtmosphere()
+
+      /* Now that we have an atmosphere, it's safe to generate the ocean */
+      generateOceans()
+    }
+  }
+
+  // RIGHT
+
+  /// A planet that exists somnewhere in the universe.
+  ///
+  /// Planets have many properties. For example, the best planets
+  /// have atmospheres and bodies of water to support life.
+  class Planet {
+    /// Terraforms the planet, by adding an atmosphere and ocean
+    /// that is hospitable for life.
+    ///  - Precondition: the planet does not already have an atmosphere.
+    ///    Adding another atmosphere to a planet that already has one is a very bad idea.
+    func terraform() {
+      // Generate the atmosphere first, before generating the ocean.
+      // Otherwise, the water will just boil off immediately.
+      generateAtmosphere()
+
+      // Now that we have an atmosphere, it's safe to generate the ocean
+      generateOceans()
+    }
+  }
+  ```
+
+  </details>
+
 ### Functions
 
 * <a id='omit-function-void-return'></a>(<a href='#omit-function-void-return'>link</a>) **Omit `Void` return types from function definitions.** [![SwiftFormat: redundantVoidReturnType](https://img.shields.io/badge/SwiftFormat-redundantVoidReturnType-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantVoidReturnType)
--- a/README.md
+++ b/README.md
@@ -925,7 +925,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   // WRONG
 
   /**
-  * A planet that exists somnewhere in the universe.
+  * A planet that exists somewhere in the universe.
   *
   * Planets have many properties. For example, the best planets
   * have atmospheres and bodies of water to support life.
@@ -951,7 +951,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   // RIGHT
 
-  /// A planet that exists somnewhere in the universe.
+  /// A planet that exists somewhere in the universe.
   ///
   /// Planets have many properties. For example, the best planets
   /// have atmospheres and bodies of water to support life.
--- a/README.md
+++ b/README.md
@@ -917,7 +917,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='comments'></a>(<a href='#comments'>link</a>) **Comment blocks should use single-line comments (`//` and `///`)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
+* <a id='comments'></a>(<a href='#comments'>link</a>) **Comment blocks should use single-line comments (`//` for code comments and `///` for documentation comments)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
 
   <details>
 
--- a/README.md
+++ b/README.md
@@ -917,7 +917,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='comments'></a>(<a href='#comments'>link</a>) **Comment blocks should use single-line comments (`//` for code comments and `///` for documentation comments)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
+* <a id='single-line-comments'></a>(<a href='#single-line-comments'>link</a>) **Comment blocks should use single-line comments (`//` for code comments and `///` for documentation comments)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
 
   <details>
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='comments'></a>(<a href='#comments'>link</a>) **Comment blocks should use single-line comments (`//` and `///`)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
```
erichoracek, line 920: Suggest making it clear when each style is used
```suggestion
* <a id='comments'></a>(<a href='#comments'>link</a>) **Comment blocks should use single-line comments (`//` for code comments and `///` for documentation comments)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
```

README.md
```diff
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+        Adding another atmosphere to a planet that already has one is a very bad idea.
+    */
+    func terraform() {
+      /* 
+      Generate the atmosphere first, before generating the ocean.
+      Otherwise, the water will just boil off immediately.
+      */
+      generateAtmosphere()
+
+      /* Now that we have an atmosphere, it's safe to generate the ocean */
```
erichoracek, line 947: Should our documentation comments wrap at our 100 column limit? I don't think we have any rule that encourages the current length that's being used

README.md
```diff
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  /// Planets have many properties. For example, the best planets
+  /// have atmospheres and bodies of water to support life.
+  class Planet {
+    /// Terraforms the planet, by adding an atmosphere and ocean
+    /// that is hospitable for life.
+    ///  - Precondition: the planet does not already have an atmosphere.
+    ///    Adding another atmosphere to a planet that already has one is a very bad idea.
+    func terraform() {
+      // Generate the atmosphere first, before generating the ocean.
+      // Otherwise, the water will just boil off immediately.
```
erichoracek, line 965: Same question re: column limit

README.md
```diff
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    /**
+      Terraforms the planet, by adding an atmosphere and ocean
+      that is hospitable for life.
+      - Precondition: the planet does not already have an atmosphere.
+        Adding another atmosphere to a planet that already has one is a very bad idea.
+    */
+    func terraform() {
+      /* 
+      Generate the atmosphere first, before generating the ocean.
+      Otherwise, the water will just boil off immediately.
```
erichoracek, line 943: 😆

erichoracek (approved):
Love it, great to hear that there's a rule to enforce this!

README.md
calda: good idea

README.md
calda: > Should our documentation comments wrap at our 100 column limit?

One thing to consider here is that I don't think there are any auto-formatting rule implementations that handle wrapping comments

>  I don't think we have any rule that encourages the current length that's being used

What do you mean?

README.md
```diff
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
 
+* <a id='comments'></a>(<a href='#comments'>link</a>) **Comment blocks should use single-line comments (`//` for code comments and `///` for documentation comments)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
+
+  <details>
+
+  ```swift
+  // WRONG
+
+  /**
+  * A planet that exists somnewhere in the universe.
```
calda, line 928:
```suggestion
  * A planet that exists somewhere in the universe.
```

README.md
```diff
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+      generateAtmosphere()
+
+      /* Now that we have an atmosphere, it's safe to generate the ocean */
+      generateOceans()
+    }
+  }
+
+  // RIGHT
+
+  /// A planet that exists somnewhere in the universe.
```
calda, line 954:
```suggestion
  /// A planet that exists somewhere in the universe.
```

README.md
erichoracek: > One thing to consider here is that I don't think there are any auto-formatting rule implementations that handle wrapping comments
> 
> > I don't think we have any rule that encourages the current length that's being used
> 
> What do you mean?

calda: Aha, I see -- this is another good question about comment style. I personally tend to manually balance comment blocks so all of the lines in a block are approximately the same length, rather than wrapping specifically at the column limit. I think I also tend to wrap comments are more like 70-80 columns by default, rather than 100 columns, which is also just a stylistic thing. I know different folks have different preferences / practices here.

I think it would be great to have consistent guidelines for comment wrapping, with auto-formatting.

calda: Not that all of the wrapping in the example stands up to much scrutiny lol! Some of these comments would be more pleasing if wrapped a bit differently

calda: Wrapping exactly at the 100 column limit looks like:

```swift
/// A planet that exists somnewhere in the universe.
///
/// Planets have many properties. For example, the best planets have atmospheres and bodies of 
/// water to support life.
class Planet {
  ///  - Precondition: the planet does not already have an atmosphere. Adding another atmosphere 
  ///    to a planet that already has one is a very bad idea.
  func terraform() {
    // Generate the atmosphere first, before generating the ocean. Otherwise, the water will just 
    // boil off immediately.
    generateAtmosphere()

    // Now that we have an atmosphere, it's safe to generate the ocean
    generateOceans()
  }
}
```

a slightly improved version of the existing wrapping could like like this:

```swift
/// A planet that exists somnewhere in the universe.
///
/// Planets have many properties. For example, the best planets
/// have atmospheres and bodies of water to support life.
class Planet {
  ///  - Precondition: the planet does not already have an atmosphere.
  ///    Adding another atmosphere to a planet that already has one is a very bad idea.
  func terraform() {
    // Generate the atmosphere first, before generating the ocean.
    // Otherwise, the water will just boil off immediately.
    generateAtmosphere()

    // Now that we have an atmosphere, it's safe to generate the ocean
    generateOceans()
  }
}
```

README.md
```diff
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='comments'></a>(<a href='#comments'>link</a>) **Comment blocks should use single-line comments (`//` for code comments and `///` for documentation comments)**, rather than multi-line comments (`/* ... */` and `/** ... */`). [![SwiftFormat: blockComments](https://img.shields.io/badge/SwiftFormat-blockComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#blockComments)
```
bachand, line 920: Since these links are meant to be permalinks, can we be more description with the anchor than "comments"? That allows us flexibility to add more comment rules in the future without ambiguity.

README.md
```diff
@@ -917,6 +917,62 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  * A planet that exists somewhere in the universe.
+  *
+  * Planets have many properties. For example, the best planets
+  * have atmospheres and bodies of water to support life.
+  */
+  class Planet {
+    /**
+      Terraforms the planet, by adding an atmosphere and ocean
+      that is hospitable for life.
+      - Precondition: the planet does not already have an atmosphere.
```
bachand, line 937: Do you think we should add a precondition to the code? We could assume it's inside of `generateAtmosphere()` though I think it would be most logical for it to live in `terraform()`.

README.md
bachand: I think it would be really useful in the future to have a rule with autocorrection around consistent comment wrapping. I try to have my comments be <= 100 lines but it can be very tedious at times with multi-line comments.

bachand: I know @swiftal64 has had thoughts about this in the past too.

bachand: Given that we say to wrap lines at 100 characters without any clarification on whether they are code lines or comments lines, we could also possibly avoid a new rule and simply add wrapping to `blockComments` in SwiftFormat.

bachand (approved):

README.md
calda: The "precondition" itself is a bit tangential to the purpose of the sample code, maybe we should simplify the comment instead to not imply the presence of an actual `precondition(...)`. Maybe something like `- Note:` would work here instead. Thoughts?

bachand: Yes I like the note idea.

README.md
erichoracek: Yeah, it would be great to have a rule that enforces comments are <100 columns and automatically reflows comments onto new lines. Hopefully that would make it so that you could just write long block comments and then have the linter take care of reflowing them, and in Cal's case he could make them shorter as well.

PR #158, opened 2022-05-02
Fix redundantRawValues example

sammygutierrez:
#### Summary

Fixes example code

#### Reasoning

Should be `redundantRawValues` instead of `redundantEnumValues`. 
https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantRawValues


calda (approved):
Thanks!

PR #164, opened 2022-06-21
Add rule to remove blank lines from the top and bottom of scopes (excluding type bodies)

calda:
#### Summary

This PR proposes a new rule to remove blank lines from the top and bottom of scopes (excluding type bodies, where we continue to permit blank lines):

```swift

// WRONG
class Planet {

  func terraform() {

    generateAtmosphere()
    generateOceans()

  }

}

// RIGHT
class Planet {

  func terraform() {
    generateAtmosphere()
    generateOceans()
  }

}

// Also fine!
class Planet {
  func terraform() {
    generateAtmosphere()
    generateOceans()
  }
}
```

#### Reasoning

Blank lines at the top and bottom of methods are unnecessary, since the indentation and braces already provide sufficient visual spacing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2185,6 +2185,44 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2185,6 +2185,44 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    
+  }
+
+  // Also fine!
+  class Planet {
+    func terraform() {
+      generateAtmosphere()
+      generateOceans()
+    }
+  }
```
erichoracek, line 2221: Would it be possible to reverse the order of these and include some `MARK`s to justify the extra newline in the case of the leading/trailing whitespace example? To my eye, the first example feels like it has redundant information; i.e. the terraform method is denoted as within the `Planet` scope both through indentation and leading/trailing newlines, which feels out of the spirit of our style guide in my interpretation

calda: Sure, I think reversing the examples makes sense -- I definitely prefer not including top/bottom blank lines in type bodies this short

PR #167, opened 2022-07-12
Include a single space in a set of empty braces (`{ }`)

calda:
#### Summary

This PR proposes a new rule to include a single space in a set of empty braces (`{ }`).

```swift
// WRONG
extension Spaceship: Trackable {}

extension SpaceshipView {
  override var accessibilityIdentifier: String {
    get { spaceship.name }
    set {}
  }
}

// RIGHT
extension Spaceship: Trackable { }

extension SpaceshipView {
  override var accessibilityIdentifier: String {
    get { spaceship.name }
    set { }
  }
}
```

#### Reasoning

We should either always or never include a space between empty braces. When testing the different options in our codebase,
 - using `--emptybraces spaced` (this PR) results in a +4500/-5300 diff across 2600 files
 - using `--emptybraces no-space` results in a +6100/-6848 diff across 3700 files

This suggests there is a slight preference for `--emptybraces spaced` in our existing codebase, since adopting it results in a smaller diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1027,6 +1027,34 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='space-in-empty-braces'></a>(<a href='#space-in-empty-braces'>link</a>) Include a single space in an empty set of braces (`{ }`). [![SwiftFormat: emptyBraces](https://img.shields.io/badge/SwiftFormat-emptyBraces-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#emptyBraces)
+
+  <details>
+
+  ```swift
+  // WRONG
+  extension Spaceship: Trackable {}
+
+  extension SpaceshipView {
+    var accessibilityIdentifier: String {
+      get { spaceship.name }
+      set {}
+    }
+  }
+
+  // RIGHT
+  extension Spaceship: Trackable { }
+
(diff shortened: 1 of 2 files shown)


bachand (approved):

julianozen:
I like that we are introducing consistency here. Great change Cal. Why 1 space instead of no space though?

calda:
I think looking at existing usage patterns is a helpful objective measure, to supplement our individual subjective style preferences. If there are 6100 instances of `{ }` and only 4500 instances of `{}` then I think that indicates that folks have a slight preference for `{ }`.

Another reason I think including a space makes sense is because its aligns with our existing _"Braces should be surrounded by a single whitespace character (either a space, or a newline) on each side."_ ([link](https://github.com/airbnb/swift#whitespace-around-braces)) and _"Closures should have a single space or newline inside each brace"_ ([link](https://github.com/airbnb/swift#closure-brace-spacing)) rules. 

One example that I think illustrates the consistency / symmetry of `{ }` nicely is this common accessibility pattern:

```swift
// With a space -- feels consistent and pleasingly symmetric
override var accessibilityIdentifier: String {
  get { spaceship.name }
  set { }
}

// Without a space -- the get and set feel formatted inconsistently
override var accessibilityIdentifier: String {
  get { spaceship.name }
  set {}
}
```

PR #168, opened 2022-07-18
Add Swift Package Manager command plugin

calda:
This PR adds a Swift Package Manager command plugin, that formats source files in Swift package targets according to our style guide.

To use this command plugin, all you have to do is add this repo as a dependency:

```swift
dependencies: [
  .package(url: "https://github.com/airbnb/swift", branch: "master"),
]
```

then run the `format` command plugin:
```
$ swift package format
```

and that's it!

> With the Xcode 13 toolchain, or in a non-interactive shell, you have to use `swift package --allow-writing-to-package-directory format` to give the tool permission to modify the package directory. I'm pretty sure that with the Xcode 14 toolchain you can omit the permission flag, and it will ask you interactively in the shell to grant permission, but I haven't tested that flow.

This seems a lot easier than the current setup that Lottie and Epoxy are using, for example -- they have to install the `mint` tool, install SwiftFormat and SwiftLint using `mint`, include copies of our config files in their repo (that quickly drift out of date), and manually invoke the two tools. Having a single command to do all of this is a nice simplification (and it will definitely make it easier to keep things up-to-date)!


AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -0,0 +1,74 @@
+# options
```
calda, line 1: To get these config files to appear in `Bundle.main` for the executable target, I had to move them from the `resources` folder to the `AirbnbSwiftFormatTool` plugin

AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -1,6 +1,3 @@
-# Current version of SwiftFormat used at Airbnb: 
-# https://github.com/calda/SwiftFormat/releases/tag/0.49.11-beta-2
```
calda, line 2: Since the SwiftFormat version has to be specified in `Package.swift`, I figure we can remove it from this file so we have a single source of truth

.github/workflows/main.yml
```diff
@@ -0,0 +1,14 @@
...
+  pull_request:
+    branches: [ master ]
+jobs:
+  build-package:
+    name: "Test Plugin"
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test Plugin
+        run: swift package --allow-writing-to-package-directory format --lint
```
calda, line 14: Added a CI step to test that the plugin build and runs correctly -- as a bonus it also lints the plugin source files 😆

calda: This is set up correctly as far as I can tell, but it isn't running on this PR. Not sure why -- maybe it needs to be merged first and then it will run on subsequent PRs?

.github/workflows/main.yml
```diff
@@ -0,0 +1,14 @@
+name: CI
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
bachand, line 6: Are we sure that specifying the branch here is correct? I don't see CI on this PR running. Here's how we did this on Swift Inspector FWIW: https://github.com/fdiaz/SwiftInspector/blob/53aec4d6fa12ba0801ca97b1368eeae60a940a59/.github/workflows/ci.yml#L3

.github/workflows/main.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  test-plugin:
+    name: "Test Plugin"
+    runs-on: macos-latest
```
bachand, line 10: I think we should specify a version so this doesn't break without a code change.

.github/workflows/main.yml
bachand: > This is set up correctly as far as I can tell, but it isn't running on this PR. Not sure why -- maybe it needs to be merged first and then it will run on subsequent PRs?

See above comment.

bachand: > Added a CI step to test that the plugin build and runs correctly -- as a bonus it also lints the plugin source files, and validates that `airbnb.swiftformat` and `swiftlint.yml` are formatted correctly (or the tool will fail to run).

Can we check if there are changes after this code runs so that we fail CI if the plugin source files don't conform to our style guide?

.gitignore
```diff
@@ -0,0 +1,4 @@
+Package.resolved
```
bachand, line 1: I think we should check this in so we have a record in source control of what versions of dependencies we're using.

Package.swift
```diff
@@ -0,0 +1,46 @@
+// swift-tools-version: 5.6
+import PackageDescription
+
+let package = Package(
+  name: "AirbnbSwiftFormatTool",
```
bachand, line 5:
```suggestion
  name: "AirbnbSwift",
```
I feel like we can take this approach to leave our options open if we want this package to do more.

AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -0,0 +1,84 @@
...
+/// based on the Airbnb Swift Style Guide
+@main
+struct AirbnbSwiftFormatTool: ParsableCommand {
+
+  // MARK: Internal
+
+  @Argument(help: "The directories to format")
+  var directories: [String]
+
+  @Option(help: "The abosolute path to a SwiftFormat binary")
```
bachand, line 14:
```suggestion
  @Option(help: "The absolute path to a SwiftFormat binary")
```

bachand: Same below

AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -0,0 +1,84 @@
...
+    let swiftFormat = Process()
+    swiftFormat.launchPath = swiftFormatPath
+    swiftFormat.arguments = arguments
+    return swiftFormat
+  }()
+
+  private lazy var swiftLint: Process = {
+    var arguments = directories + [
+      "--config", swiftLintConfig,
+      "--in-process-sourcekit",
```
bachand, line 67: Lazy q: what does this do and why do we specify it?

AirbnbSwiftFormatPlugin/AirbnbSwiftFormatPlugin.swift
```diff
@@ -0,0 +1,36 @@
...
+    var processArguments = context.package.targets.map { $0.directory.string } + [
+      "--swift-format-path",
+      try context.tool(named: "swiftformat").path.string,
+      "--swift-lint-path",
+      try context.tool(named: "swiftlint").path.string,
+      // The process we spawn doesn't have read/write access to the default
+      // cache file locations, so we pass in our own cache paths from
+      // the plugin's work directory.
+      "--swift-format-cache-path",
+      context.pluginWorkDirectory.string + "/swiftformat.cache",
```
bachand, line 22: What does this working directory look like? I want to see if it persists between plugin runs.

AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -0,0 +1,84 @@
...
+  var lint = false
+
+  @Option(help: "The absolute path to the SwiftFormat config file")
+  var swiftFormatConfig = Bundle.module.path(forResource: "airbnb", ofType: "swiftformat")!
+
+  @Option(help: "The absolute path to the SwiftLint config file")
+  var swiftLintConfig = Bundle.module.path(forResource: "swiftlint", ofType: "yml")!
+
+  mutating func run() throws {
+    try swiftFormat.run()
```
bachand, line 36: I wondered if instead wanted to link to the code for SwiftFormat and SwiftLint but executing the binaries is probably more robust.

bachand (approved):
Nice!

bachand:
Oh also, a GIF of this in action would be great!

.github/workflows/main.yml
calda: This is the same configuration we use for Lottie and Epoxy: https://github.com/airbnb/lottie-ios/blob/master/.github/workflows/main.yml#L3

Will try what you have for SwiftInspector though, thanks for the suggestion.

.github/workflows/main.yml
calda: Ideally the tool will emit an error exit code if `--lint` detects any issues, which would cause the CI job to fail. Will double-check that this actually works, once I get CI to run.

.gitignore
calda: Sure

AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: Will add a comment -- this is required due to SPM plugin sandboxing, I got this tip from https://theswiftdev.com/beginners-guide-to-swift-package-manager-command-plugins/

Package.swift
calda: Good idea

AirbnbSwiftFormatPlugin/AirbnbSwiftFormatPlugin.swift
calda: The documentation says:
> The plugin is in complete control of what is written under this di- rectory, and the contents are preserved between builds.

On my machine this folder is `~/Documents/swift-style-guide/.build/plugins/AirbnbSwiftFormatPlugin/outputs`. The only two files inside it are the `swiftformat.cache` and `swiftlint.cache` files we create here.

.github/workflows/main.yml
calda: With the changes in https://github.com/airbnb/swift/pull/168/commits/4408b6c4ffb904798477d26aad4a9b6efdc301f8 the tool / plugin emit a non-zero exit code when encountering a lint failure

calda: Tested this in #170, which added some intentional lint failures. The checks ran and then failed as expected: https://github.com/airbnb/swift/runs/7421227671?check_suite_focus=true

.github/workflows/main.yml
calda: Worked correctly after I merged this PR: https://github.com/airbnb/swift/pull/168#discussion_r925104433

AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -72,7 +73,7 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
     }
 
     if !lint {
-      arguments += ["--autocorrect"]
+      arguments += ["--fix"]
```
bachand, line 87: The docs say the options `--fix` and `--autocorrect` are the same. Did you notice a difference?
Screen Shot 2022-07-20 at 5 39 28 PM

calda: I tried both and didn't notice a difference

PR #169, opened 2022-07-19
Update Swift version to 5.6

calda:
This PR updates our `airbnb.swiftformat` version to Swift 5.6, which is the language version we are using at AIrbnb.

It would be nice if SwiftFormat supported a `--swiftversion latest` flag or something, to always use the most recent language version, so we don't have to keep updating this. Will see what Nick thinks about that idea.


AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -1,4 +1,5 @@
 # options
+--swiftversion 5.6
```
calda, line 2: Nick pointed out that `--swiftversion 999` should work as "always use the latest swift version". Thoughts?

fdiaz (approved):

PR #170, opened 2022-07-20
Sample change to cause lint failure in CI

calda:
This PR is a sample change that should cause the CI job to fail


calda:
CI failed, nice!

fdiaz:
Good stuff!

bachand:
Nice! @calda do we feel confident we'll fail for rules validated by both SwiftFormat and SwiftLint?

calda:
Yes! I tested both types of failures locally.

PR #172, opened 2022-07-21
Aesthetics for new Swift command plugin

bachand:
#### Summary

Cal added a command plugin to format Swift code per our style guide in https://github.com/airbnb/swift/pull/168. After watching the WWDC22 talks related to Swift package plugins I had a couple of ideas for improving the aesthetics of this plugin.

#### Reasoning

Let's avoid command naming collisions in the CLI and simplify the name of our command in Xcode.

#### Testing

I made edits to the package that didn't conform to our style guide and then ran `swift package airbnb-format-swift` and saw those edits get fixed.

@calda @airbnb/swift-styleguide-maintainers 


Package.swift
```diff
@@ -18,10 +18,10 @@ let package = Package(
   ],
   targets: [
     .plugin(
-      name: "AirbnbSwiftFormatPlugin",
+      name: "AirbnbSwiftFormat",
       capability: .command(
         intent: .custom(
-          verb: "format",
+          verb: "airbnb-swift-format",
```
bachand, line 24: When another package consumes this plugin (e.g. https://github.com/airbnb/lottie-ios/blob/97c05f8272fc8e3a2cefb3d41c6bcb90c6d28739/Rakefile#L108) it uses this name. Let's pick a more specific name to avoid collisions.

Package.swift
```diff
@@ -6,7 +6,7 @@ let package = Package(
   platforms: [.macOS(.v10_13)],
   products: [
     .executable(name: "AirbnbSwiftFormatTool", targets: ["AirbnbSwiftFormatTool"]),
-    .plugin(name: "AirbnbSwiftFormatPlugin", targets: ["AirbnbSwiftFormatPlugin"]),
+    .plugin(name: "FormatSwift", targets: ["FormatSwift"]),
```
bachand, line 9: Xcode makes clear that this plugin is from our package, so we can have a simpler name.
Screen Shot 2022-07-21 at 12 13 06 PM

Package.swift
```diff
@@ -30,14 +30,12 @@ let package = Package(
         "AirbnbSwiftFormatTool",
         .product(name: "swiftformat", package: "SwiftFormat"),
         .product(name: "swiftlint", package: "SwiftLint"),
-      ],
-      path: "AirbnbSwiftFormatPlugin"),
```
bachand, line 34: We can remove this path by putting the source in a folder under the default Plugins directory.

Package.swift
```diff
@@ -30,14 +30,12 @@ let package = Package(
...
+      ]),
     .executableTarget(
       name: "AirbnbSwiftFormatTool",
       dependencies: [
         .product(name: "ArgumentParser", package: "swift-argument-parser"),
       ],
-      path: "AirbnbSwiftFormatTool",
```
bachand, line 40: We can remove this path by putting the source in a folder under the default Sources directory.

Package.swift
```diff
@@ -18,10 +18,10 @@ let package = Package(
   ],
   targets: [
     .plugin(
-      name: "AirbnbSwiftFormatPlugin",
+      name: "FormatSwift",
       capability: .command(
         intent: .custom(
-          verb: "format",
+          verb: "airbnb-format-swift",
```
bachand, line 24: When another package consumes this plugin (e.g. https://github.com/airbnb/lottie-ios/blob/97c05f8272fc8e3a2cefb3d41c6bcb90c6d28739/Rakefile#L108) it uses this name. Let's pick a more specific name to avoid collisions. I'm not sure what happens when the names collide.

bachand, line 24: This change will break the Lottie repo since the repo depends on the master branch of this repo. I will put up a PR to Lottie before merging this change.

Package.swift
```diff
@@ -6,7 +6,7 @@ let package = Package(
   platforms: [.macOS(.v10_13)],
   products: [
     .executable(name: "AirbnbSwiftFormatTool", targets: ["AirbnbSwiftFormatTool"]),
```
bachand, line 8: I wondered if we should export this tool or simplify export the plugin for now. I also wondered if we should simplify the name of the tool but I left it as-is.

Package.swift
calda: Lottie is pinned to a specific commit in its `Package.resolved` right?

calda: I think `swift package format` is noticeably more ergonomic that `swift package airbnb-format-swift`. I'm not sure how worried we need to be about plugin command conflicts -- perhaps we can hold off on trying to work around that unless it comes up in the future?

fdiaz: I tend to agree with Cal here, I really dislike using `-` or `_` on the CLI since it's much more inconvenient to type.

fdiaz: We should likely start using releases in this repo too though once we make sure this plugin works effectively.

bachand: I'm following the [example](https://developer.apple.com/videos/play/wwdc2022/110401/) of the WWDC talk on making these plugins.

bachand: > I'm not sure how worried we need to be about plugin command conflicts -- perhaps we can hold off on trying to work around that unless it comes up in the future?

bachand: Interesting, it looks like we may not need to worry about conflicts ([source](https://github.com/apple/swift-evolution/blob/main/proposals/0332-swiftpm-command-plugins.md)).
Screen Shot 2022-07-21 at 1 37 07 PM

calda: aha cool! So I guess it would be `swift package AirbnbSwift:format` in that case.

bachand: Good call @calda on the `Package.resolved` 👍  I'm glad we are checking those files in. In any case, I've removed the change to this `verb` so it's a moot point. I agree with @fdiaz that we should start adding releases to this repo, and a changelog.

calda: definitely way better than `swift package AirbnbSwiftFormatPlugin:format`, that was a good suggestion

calda: Sounds good -- I suppose we'll make a new release when we make config changes from adding a new rule to the style guide?

calda (approved):

Package.swift
bachand: > Sounds good -- I suppose we'll make a new release when we make config changes from adding a new rule to the style guide?

👍

calda: To double check we understand how this works, I tested the Lottie repo after this change. Running `swift package format` used the previous commit as specified in `Package.resolved`. When I run `swift package update`, it updates `Package.resolved` to reference the most recent commit. I do agree that we should start cutting releases for this repo once we stabilize the tool though.

bachand: Thanks for checking @calda

PR #173, opened 2022-07-21
Improve error handling, add support for `--exclude` flag

calda:
This PR improves error handling for the format plugin, and adds support for an `--exclude` flag.

I'm adopting the plugin in the Lottie and Epoxy repos, and noticed that it was outputting a `lintFailure` even in cases where SwiftLint / SwiftFormat were failing due to reasons unrelated to the lint configuration. Now we clarify between an expected failure (a `lintFailure`) and an unknown / unexpected failure that we cannot recover from.

Epoxy specifically doesn't lint its `Tests` subdirectory, so we needed a way to configure the directories that are formatted / linted by the plugin. I figured we can use an `--exclude` flag for this, to compliment the exclude configurations in `airbnb.swiftformat` / `swiftlint.yml`.

### DIFF
--- a/Plugins/FormatSwift/Plugin.swift
+++ b/Plugins/FormatSwift/Plugin.swift
@@ -14,7 +14,14 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
     let process = Process()
     process.launchPath = try context.tool(named: "AirbnbSwiftFormatTool").path.string
 
-    var processArguments = try inputPaths(context: context) + [
+    var argumentExtractor = ArgumentExtractor(arguments)
+
+    let inputPaths = try inputPaths(
+      context: context,
+      // Consumers can exclude additional subdirectories with an `--exclude` flag
+      excluding: argumentExtractor.extractOption(named: "exclude"))
+
+    var processArguments = inputPaths + [
       "--swift-format-path",
       try context.tool(named: "swiftformat").path.string,
       "--swift-lint-path",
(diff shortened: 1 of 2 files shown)


AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -28,6 +28,9 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
   @Flag(help: "When true, source files are not reformatted")
   var lint = false
 
+  @Flag(help: "When true, logs the commands that are executed")
+  var log = false
```
calda, line 32: This was useful for diagnosing some weird SwiftLint failures I was hitting (some specific files in the Epoxy repo cause it to fail with `Bus error: 10` or `Illegal instruction: 4` for some reason).

Plugins/FormatSwift/Plugin.swift
```diff
@@ -53,21 +64,34 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
...
+      }
   }
 
 }
 
-// MARK: - LintError
+// MARK: - Error
 
-enum LintError: Error {
+enum Error: Swift.Error {
```
erichoracek, line 94: nit: feels kinda confusing to name this the same as Swift.Error especially since it isn't nested

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -41,11 +44,25 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
-      throw LintError.lintFailure
+    if log {
+      log(swiftFormat.shellCommand)
+      log(swiftLint.shellCommand)
+      log("SwiftFormat ended with exit code \(swiftFormat.terminationStatus)")
+      log("SwiftLint ended with exit code \(swiftLint.terminationStatus)")
+    }
+
+    // SwiftFormat ends with exit code 1 on lint failure, and SwiftLint ends with exit code 2
+    if swiftFormat.terminationStatus == 1 || swiftLint.terminationStatus == 2 {
```
erichoracek, line 55: nit: we could make some enums for tool error statuses to make this self-documenting
```swift
    if swiftFormat.terminationStatus == SwiftFormatTerminationStatus.lintFailue {
```

erichoracek (approved):
LGTM!

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: Good idea!

Plugins/FormatSwift/Plugin.swift
calda: haha sure I'll rename this to `CommandError` or something

PR #175, opened 2022-07-22
Add support for Xcode 14 target selection dialog

calda:
This PR adds support for the target selection dialog in Xcode 14:

image

Xcode passes in `--target` option arguments for each of the targets you select. Handing this in our existing `CommandPlugin` seems to be all we need to do to fully support Xcode 14.

Depends on #173.

### DIFF
--- a/Plugins/FormatSwift/Plugin.swift
+++ b/Plugins/FormatSwift/Plugin.swift
@@ -16,10 +16,27 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
 
     var argumentExtractor = ArgumentExtractor(arguments)
 
-    let inputPaths = try inputPaths(
-      context: context,
-      // Consumers can exclude additional subdirectories with an `--exclude` flag
-      excluding: argumentExtractor.extractOption(named: "exclude"))
+    // When ran from Xcode, the plugin command is invoked with `--target` arguments,
+    // specifying the targets selected in the plugin dialog.
+    let inputTargets = argumentExtractor.extractOption(named: "target")
+
+    var inputPaths: [String]
+    if !inputTargets.isEmpty {
+      // If a set of input targets were given, lint/format the directory for each of them
+      inputPaths = try context.package.targets(named: inputTargets).map { $0.directory.string }
+    } else {
+      // Otherwise if no targets we listed we default to linting/formatting
+      // the entire package directory.
+      inputPaths = try self.inputPaths(for: context.package)
+    }
+
+    // Filter out any excluded paths passed in with `--exclude`
+    let excludedPaths = argumentExtractor.extractOption(named: "exclude")
+    inputPaths = inputPaths.filter { path in
+      !excludedPaths.contains(where: { excludedPath in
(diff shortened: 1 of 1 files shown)


erichoracek (commented):
LGTM!

erichoracek (approved):

Plugins/FormatSwift/Plugin.swift
```diff
@@ -16,10 +16,27 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
 
     var argumentExtractor = ArgumentExtractor(arguments)
 
-    let inputPaths = try inputPaths(
-      context: context,
-      // Consumers can exclude additional subdirectories with an `--exclude` flag
-      excluding: argumentExtractor.extractOption(named: "exclude"))
+    // When ran from Xcode, the plugin command is invoked with `--target` arguments,
+    // specifying the targets selected in the plugin dialog.
+    let inputTargets = argumentExtractor.extractOption(named: "target")
```
bachand, line 21: Are targets a list of relative paths?

bachand (commented):
Nice! When I ran the plugin in Xcode 14 I saw this error.

image


I'm surprised that simply supporting `--targets` resolves the error. Did you not see this one? I assumed we needed to conform to `XcodeCommandPlugin`.

calda:
@bachand, what version of Xcode 14 are you using? This works for me on Xcode 14 beta 3 -- I never encountered the issue you're getting.

Plugins/FormatSwift/Plugin.swift
calda: I think they're the names of targets as defined in the `Package.swift`, not paths or direcories

bachand:
I was using Xcode 14 beta 3. I will try again tomorrow.

bachand:
@calda I run into the same issue. Here's the steps I followed.
1. Select Xcode 14 Beta 3 as your Xcode.
2. Create a new test app.
3. Make changes to one of the files that violate the style guide.
4. Add the package via File -> Add Packages...
5. When the package is loaded right click on the application in the project navigation and select the plugin.

Screen Shot 2022-07-27 at 9 41 08 AM




7. Run the plugin on the whole app.
Screen Shot 2022-07-27 at 9 41 31 AM



8. Allow the plugin to changes files.
Screen Shot 2022-07-27 at 9 41 56 AM

9. Notice the failure in the logs, and notice that the Swift code that you edited is not formatted per the style guide.
Screen Shot 2022-07-27 at 9 43 00 AM
Screen Shot 2022-07-27 at 9 45 29 AM


I also see a failure when compiling SwiftFormat which I always see when compiling in Xcode, but I think that's unrelated to what we're discussing.
Screen Shot 2022-07-27 at 9 43 30 AM

bachand:
I wonder if your testing worked since we were running the plugin on a SPM package and not arbitrary Xcode targets?

calda:
Hmmm yes maybe that explains it, I was only testing in Xcode workspaces directly opened from a `Package.swift` file.

PR #177, opened 2022-07-23
Add files argument to only lint specific paths

jayhickey:
#### Summary

This adds a `--files` parameter to the Swift package plugin so that you can choose to _only_ format specific files that you wish rather than entire targets or packages.

#### Reasoning

In some situations it's useful to only format specific files. For example you might want to set up a pre-commit git hook to format _only_ staged swift files before committing them. 

### DIFF
--- a/Plugins/FormatSwift/Plugin.swift
+++ b/Plugins/FormatSwift/Plugin.swift
@@ -21,19 +21,17 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
     let inputTargets = argumentExtractor.extractOption(named: "target")
 
     // Only lint the paths passed to `--paths`
-    let strictPaths = argumentExtractor.extractOption(named: "paths")
+    let paths = argumentExtractor.extractOption(named: "paths")
 
-    var inputPaths: [String]
+    var inputPaths: [String] = paths
     if !inputTargets.isEmpty {
       // If a set of input targets were given, lint/format the directory for each of them
-      inputPaths = try context.package.targets(named: inputTargets).map { $0.directory.string }
+      inputPaths += try context.package.targets(named: inputTargets).map { $0.directory.string }
     } else if strictPaths.isEmpty {
       // Otherwise if no targets or paths listed we default to linting/formatting
       // the entire package directory.
       inputPaths = try self.inputPaths(for: context.package)
     }
-
-    inputPaths += strictPaths
 
     // Filter out any excluded paths passed in with `--exclude`
     let excludedPaths = argumentExtractor.extractOption(named: "exclude")
--- a/Plugins/FormatSwift/Plugin.swift
+++ b/Plugins/FormatSwift/Plugin.swift
@@ -20,10 +20,9 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
     // specifying the targets selected in the plugin dialog.
     let inputTargets = argumentExtractor.extractOption(named: "target")
 
-    // Only lint the paths passed to `--paths`
-    let paths = argumentExtractor.extractOption(named: "paths")
+    // If given, lint only the paths passed to `--paths`
+    var inputPaths = argumentExtractor.extractOption(named: "paths")
 
-    var inputPaths: [String] = paths
     if !inputTargets.isEmpty {
       // If a set of input targets were given, lint/format the directory for each of them
       inputPaths += try context.package.targets(named: inputTargets).map { $0.directory.string }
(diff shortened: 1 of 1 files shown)


Plugins/FormatSwift/Plugin.swift
```diff
@@ -20,12 +20,15 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
...
     let inputTargets = argumentExtractor.extractOption(named: "target")
 
-    var inputPaths: [String]
+    // Only lint the files passed to `--files`
+    let inputFiles = argumentExtractor.extractOption(named: "files")
+
+    var inputPaths: [String] = inputFiles
     if !inputTargets.isEmpty {
       // If a set of input targets were given, lint/format the directory for each of them
       inputPaths = try context.package.targets(named: inputTargets).map { $0.directory.string }
```
calda, line 29: Should we support a combination of both `--targets` and `--files`? Right now it looks like if you provide both then we ignore the `--files` inputs. Otherwise if we want to explicitly only support one or the other, I think we should probably throw an error when both are given. Thoughts?

calda (commented):
Thanks for the contribution!

Plugins/FormatSwift/Plugin.swift
```diff
@@ -20,12 +20,15 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
     // specifying the targets selected in the plugin dialog.
     let inputTargets = argumentExtractor.extractOption(named: "target")
 
-    var inputPaths: [String]
+    // Only lint the files passed to `--files`
+    let inputFiles = argumentExtractor.extractOption(named: "files")
```
calda, line 24: This supports both files and directories, so maybe we should name the option `--paths` instead. Thoughts?

Plugins/FormatSwift/Plugin.swift
jayhickey: Ah yes good catch, I agree! I'll change it to support both.

Plugins/FormatSwift/Plugin.swift
```diff
@@ -20,16 +20,21 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
     // specifying the targets selected in the plugin dialog.
     let inputTargets = argumentExtractor.extractOption(named: "target")
 
+    // Only lint the paths passed to `--paths`
```
calda, line 23:
```suggestion
    // Only lint the paths passed to `--paths` if given
```

Plugins/FormatSwift/Plugin.swift
```diff
@@ -20,16 +20,21 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
...
+    let strictPaths = argumentExtractor.extractOption(named: "paths")
```
calda, line 24: Having a single `inputPaths` property and then doing `inputPaths += targetPaths` below may be a bit more idiomatic

```suggestion
    var inputPaths = argumentExtractor.extractOption(named: "paths")
```

calda (approved):

bachand:
One thought... is it possible for SPM command plugins to accept arguments in addition to options? I wonder if instead of passing `--files` the arguments you pass could be files or directories to lint.

So instead of

```bash
swift package format --paths Sources Tests Package.swift
```

you'd pass

```bash
swift package format Sources Tests Package.swift
```

There are pros/cons to each approach though it's pretty idiomatic for CLI tools to take paths as arguments.

jayhickey:
> One thought... is it possible for SPM command plugins to accept arguments in addition to options? I wonder if instead of passing `--files` the arguments you pass could be files or directories to lint.

You can retrieve those additional arguments via `argumentExtractor.remainingArguments`, that is already being done for the purpose of passing remaining arguments down to the child process [here](https://github.com/airbnb/swift/blob/1d8bb91f317a5f48be9a26e77558cae7398da385/Plugins/FormatSwift/Plugin.swift#L57-L58):

```
    // Pass any remaining arguments directly to the child process
    processArguments += argumentExtractor.remainingArguments
```

The challenge with your approach (which I agree is more idiomatic) is that those other arguments could be arguments for the child swiftlint / swiftformat processes. Without the flag we don't have a good way to distinguish between those args being files / directories for linting vs additional args for the child processes.

bachand:
Thanks for the explanation @jayhickey . That makes sense. I guess it could be a pain for the plugin to be aware of all options for the underlying CLI tool.

PR #180, opened 2022-07-27
Use `swift-tools-version` from `Package.swift` as minimum Swift version by default

calda:
This PR updates the package plugin to use the `swift-tools-version` specified in `Package.swift` as the minimum Swift version by default. This can also be customized manually by passing in a `--swift-version` option.

Previously the SwiftFormat `--swiftversion` was included in `airbnb.swiftformat` and was not overridable. We listed 5.6, but Lottie for example has a minimum supported Swift version of 5.3. To override the default `--swiftversion 5.6` option, I set up a `.swift-version` file at the repo root to specify `5.3`. This file is conflicting with Cocoapods (airbnb/lottie-ios#1672), so turned out to not be a good approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,10 @@ $ swift package format --exclude Tests
 # Alternatively you can explicitly list the set of paths and/or SPM targets:
 $ swift package format --paths Sources Tests Package.swift
 $ swift package format --targets AirbnbSwiftFormatTool
+
+# The plugin infers your package's minimum Swift version from the `swift-tools-version`
+# in your `Package.swift`, but you can provide a custom value with `--swift-version`:
+$ swift package format --swift-version 5.3
 ```
 
 </details>
--- a/Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
+++ b/Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
@@ -37,6 +37,9 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
   @Option(help: "The absolute path to the SwiftLint config file")
   var swiftLintConfig = Bundle.module.path(forResource: "swiftlint", ofType: "yml")!
 
+  @Option(help: "The project's minimum Swift version")
+  var swiftVersion: String?
+
   mutating func run() throws {
     try swiftFormat.run()
     swiftFormat.waitUntilExit()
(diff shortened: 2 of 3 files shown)


erichoracek (approved):
LGTM

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -83,6 +86,10 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
       arguments += ["--lint"]
     }
 
+    if let swiftVersion = swiftVersion {
+      arguments += ["--swiftversion", swiftVersion]
```
bachand, line 90: Will this take precedence over https://github.com/airbnb/swift/blob/1bf961396cc9c690db37936bfbdc8b5f29e844af/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat#L2?

README.md
```diff
@@ -59,6 +59,10 @@ $ swift package format --exclude Tests
 # Alternatively you can explicitly list the set of paths and/or SPM targets:
 $ swift package format --paths Sources Tests Package.swift
 $ swift package format --targets AirbnbSwiftFormatTool
+
+# The plugin infers your package's minimum Swift version from the `swift-tools-version`
+# in your `Package.swift`, but you can provide a custom value with `--swift-version`:
+$ swift package format --swift-version 5.3
```
bachand, line 65: Are we sure this works? In our code above we are _always_ creating `--swift-version` option and passing that to the underlying tool. If someone executed this command and their Package.swift had a tools version of 5.6 I think we'd pass `--swift-version 5.3 --swift-version 5.6` to the tool.

bachand: It seems like we should avoid constructing a default option for the underlying tool if the user specified `--swift-version` manually.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: Yes, it does

README.md
calda: This command behaves as expected. I suppose the plugin is passing e.g. `--swift-version 5.6 --swift-version 5.3` to the underlying tool, and since this option isn't array Swift Argument Parser is selecting the last value. That sounds like its potentially relying on UB, so I agree that we should avoid passing default option if the user also provides one. Will post a PR.

PR #181, opened 2022-07-27
Add `XcodeCommandPlugin` to fully support Xcode 14 

calda:
This PR adds an `XcodeCommandPlugin` to fully support Xcode 14.

https://user-images.githubusercontent.com/1811727/181390333-36e120ad-b61c-44d4-b3f7-23de20504719.mov

#175 added support for invoking the SPM `CommandPlugin` _on a Swift Package_ when opened in Xcode 14. 

This didn't support other types of Xcode workspaces, e.g. an iOS app project defined in a `.xcodeproj`, as @bachand pointed out in https://github.com/airbnb/swift/pull/175#issuecomment-1197008468. To support that use case, we have to add an `XcodeCommandPlugin` conformance to our plugin.

I also started running in to the `Cannot import "swiftformat" as "SwiftFormat"` error that @bachand mentioned, which was blocking me from testing this, so I updated the plugin to use precompiled binary executables instead of compiling SwiftFormat and SwiftLint from source. This significantly reduces how long it takes to run the plugin and simplifies our dependencies, which are all nice wins.


Package.swift
```diff
@@ -41,4 +35,15 @@ let package = Package(
         .process("airbnb.swiftformat"),
         .process("swiftlint.yml"),
       ]),
+
+    // We don't always use official SwiftFormat releases from the nicklockwood/SwiftFormat
+    // repo, so we host our own artifactbundle instead of downloading one.
+    .binaryTarget(
+      name: "SwiftFormat",
+      path: "resources/SwiftFormat.artifactbundle"),
```
erichoracek, line 43: Can we host this on your SwiftFormat fork? It seems like we probably don't want to check this in since it's a 6 MB executable

calda: Sure, I can zip it and attach it to the release for this version

calda: Updated in https://github.com/airbnb/swift/pull/181/commits/bad3bcb1d47eb0cddd8aaf44b814a03dafd8b2b6

Package.swift
```diff
@@ -8,13 +8,6 @@ let package = Package(
     .plugin(name: "FormatSwift", targets: ["FormatSwift"]),
   ],
   dependencies: [
-    .package(url: "https://github.com/calda/SwiftFormat", exact: "0.49.11-beta-2"),
-    // The `SwiftLintFramework` target uses "unsafe build flags" so Xcode doesn't
-    // allow us to reference a specific version number. To work around that,
-    // we can reference the specific commit for that version.
```
calda, line 14: This was actually also blocking consumers from specifying version numbers like `from: 1.0.0`, so that was another reason why we needed to move to prebuilt binary executables.

Plugins/FormatSwift/Plugin.swift
```diff
@@ -96,6 +115,32 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
...
+
+  func performCommand(context: XcodePluginContext, arguments: [String]) throws {
+    var argumentExtractor = ArgumentExtractor(arguments)
+
+    // When ran from Xcode, the plugin command is invoked with `--target` arguments,
+    // specifying the targets selected in the plugin dialog.
+    //  - Unlike SPM targets which are just directories, Xcode targets are
+    //    an arbitrary collection of paths.
+    let inputTargetNames = Set(argumentExtractor.extractOption(named: "target"))
+    let inputPaths = context.xcodeProject.targets.lazy
```
erichoracek, line 129: Why make this `lazy` if we're just converting it to an array below?

Package.swift
```diff
@@ -8,13 +8,6 @@ let package = Package(
...
-    //  - This is top-of-tree master from 7/22/22, because the most recent release version
-    //    (0.47.1) seems to have issues in some real-world projects like airbnb/epoxy-ios.
-    .package(url: "https://github.com/realm/SwiftLint", revision: "c5aa806"),
```
erichoracek, line 17: Will this allow us to not have the full source of these two tools and all of the deps in projects that consume this plugin? I was noticing that Epoxy had [way more resolved package deps](https://github.com/airbnb/epoxy-ios/blob/9b9b1d3312dc2daaaeb06346b72451e8accfe7dd/Epoxy.xcworkspace/xcshareddata/swiftpm/Package.resolved#L40-L120) after we added this dep

erichoracek (commented):
Nice, LGTM!

Package.swift
calda: Yes! The only remaining source dependency is Swift Argument Parser, everything else is precompiled.

Plugins/FormatSwift/Plugin.swift
calda: I'd expect `lazy` here to improve performance a bit for large targets with lots of files by reducing the total number of loop iterations (e.g. not having to do two separate loop passes for the `map` and then `filter`). Could be mistaken though.

erichoracek (approved):

Plugins/FormatSwift/Plugin.swift
erichoracek: Got it, makes sense, so we get better data locality since each of the operations for a single element occur in sequence?

calda: Yeah I think so

calda: Maybe it doesn't really matter / is a micro-optimization, I didn't instrument it of course. I just tend to lean for `lazy` in cases where a functional chain on a collection has a bunch of different steps like this.

Plugins/FormatSwift/Plugin.swift
```diff
@@ -96,6 +115,32 @@ struct AirbnbSwiftFormatPlugin: CommandPlugin {
 
 }
 
+#if canImport(XcodeProjectPlugin)
+extension AirbnbSwiftFormatPlugin: XcodeCommandPlugin {
+
+  func performCommand(context: XcodePluginContext, arguments: [String]) throws {
```
bachand, line 121: Nice work @calda ! I see that in our current implementation we are only considering the arguments in the standard plugin implementation.

In Xcode 14 it's possible to pass arguments as well. Can we honor those arguments passed in from Xcode for completeness?

Screen Shot 2022-07-28 at 6 03 11 PM

bachand (commented):
The binary dependencies are great!

Plugins/FormatSwift/Plugin.swift
calda: I think it sort of makes sense for the SPM package plugin and Xcode plugin to have different sets of arguments, especially considering the supported functionality is different (e.g. there isn't a feasible implementation of `--paths` for the Xcode plugin). Also since you have to add the arguments manually each time you invoke the tool in Xcode, I somewhat assume excluding individual files wouldn't be a common use case. It wouldn't hurt to support it though.

bachand: I see what you're saying @calda . I mostly want to make sure we're intentional.

> there isn't a feasible implementation of `--paths` for the Xcode plugin

Can you elaborate on why this isn't possible? Asking to learn.

calda: I was thinking of the difference between SPM targets and Xcode targets -- SPM targets correspond directly to folders, so have a `directory` path for these input paths to be resolved against. On the other hand Xcode targets are just a collection of files so don't have a `directory` path. The Xcode project itself does have a `directory`, though, now that I look at it, so you could resolve paths against that directory.

bachand: Makes sense. I think it's reasonable to ignore `--files` when within Xcode.

PR #182, opened 2022-07-29
Improve handling of `--swift-format` and `--exclude` arguments

calda:
This PR makes changes changes based on feedback from @bachand:
 - Updated plugin to not pass multiple `--swift-format` arguments to underlying tool (https://github.com/airbnb/swift/pull/180#discussion_r932780947)
 - Updated `XcodeCommandPlugin` to support `--exclude` arguments (https://github.com/airbnb/swift/pull/181#discussion_r932785998)

### DIFF
--- a/Plugins/FormatSwift/Plugin.swift
+++ b/Plugins/FormatSwift/Plugin.swift
@@ -16,13 +16,21 @@ struct AirbnbSwiftFormatPlugin {
   func performCommand(
     context: CommandContext,
     inputPaths: [String],
-    additionalArguments: [String])
+    arguments: [String])
     throws
   {
-    let process = Process()
-    process.launchPath = try context.tool(named: "AirbnbSwiftFormatTool").path.string
+    var argumentExtractor = ArgumentExtractor(arguments)
 
-    process.arguments = inputPaths + [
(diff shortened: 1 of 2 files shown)


Plugins/FormatSwift/Plugin.swift
```diff
@@ -34,8 +42,16 @@ struct AirbnbSwiftFormatPlugin {
       context.pluginWorkDirectory.string + "/swiftformat.cache",
       "--swift-lint-cache-path",
       context.pluginWorkDirectory.string + "/swiftlint.cache",
-    ] + additionalArguments
+    ] + argumentExtractor.remainingArguments
 
+    if arguments.contains("--log") {
+      // swiftlint:disable:next no_direct_standard_out_logs
+      print(launchPath, arguments.joined(separator: " "))
```
bachand, line 49: Since we are logging in the plugin and in the underlying tool, can we prefix each log with where it comes from so the log file is easier to parse?

bachand (approved):

Plugins/FormatSwift/Plugin.swift
calda: Good idea

PR #183, opened 2022-08-04
Include binary targets directly in repo to work around Xcode performance issue

calda:
This PR fixes an issue where Xcode projects with this package as a dependency would spin indefinitely at 100% CPU load:

Screen Shot 2022-08-04 at 6 49 23 AM

This was happening because Xcode was constantly re-resolving this package's remote binary dependencies:

https://user-images.githubusercontent.com/1811727/182863964-2d05ebc2-e749-4d26-a066-debeacdb5a0f.mov

This happens in both Xcode 13.4 and Xcode 14 beta 3/4. I'll file a feedback with a repro case so hopefully this is fixed in the future. 

As a workaround in the meantime, we can prevent this from happening by including the binary dependencies directly in this repo instead of specifying remote dependencies that have to be downloaded separately.


calda:
Filed FB11062421 for this issue.

calda:
It's not ideal that this requires adding 70MB+ of binary dependencies to this repo's git history. Closing in favor of workarounds like https://github.com/airbnb/lottie-ios/pull/1705 while we wait for this issue to get fixed in a future version of Xcode 🤞🏻

PR #192, opened 2022-09-12
Add rule to prefer shorthand `if let` optional unwrapping syntax

calda:
#### Summary

This PR proposes a new rule to prefer the shorthand `if let` optional unwrapping syntax added in Swift 5.7:

> Omit the right-hand side of the expression when unwrapping an optional property to a non-optional property with the same name.

```swift
// WRONG
if
  let galaxy = galaxy,
  galaxy.name == "Milky Way"
{ … }

guard
  let galaxy = galaxy,
  galaxy.name == "Milky Way"
else { … }

// RIGHT
if
  let galaxy,
  galaxy.name == "Milky Way"
{ … }

guard
  let galaxy,
  galaxy.name == "Milky Way"
else { … }
```

#### Reasoning

We are adopting Swift 5.7 soon, so we should adopt guidance for its new language features.

As discussed in the Motivation section of [SE-0345](https://github.com/apple/swift-evolution/blob/main/proposals/0345-if-let-shorthand.md), the new shorthand optional unwrapping syntax removes unnecessary boilerplate while retaining clarity. Auto-formatting for this rule is implemented in https://github.com/nicklockwood/SwiftFormat/pull/1165.

### DIFF
--- a/README.md
+++ b/README.md
@@ -413,7 +413,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     func request(completion: () -> Void) {
       API.request() { [weak self] response in
-        guard let self = self else { return }
+        guard let self else { return }
         // Do work
         completion()
       }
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -729,6 +729,38 @@ _You can enable the following settings in Xcode by running [this script](resourc
     & CivilizationServiceProviding
   ```
 
+* <a id='if-let-shorthand'></a>(<a href='#if-let-shorthand'>link</a>) Omit the right-hand side of the expression when unwrapping an optional property to a non-optional property with the same name. [![SwiftFormat: redundantOptionalBinding](https://img.shields.io/badge/SwiftFormat-redundantOptionalBinding-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantOptionalBinding)
```
bachand, line 732:
```suggestion
* <a id='prefer-if-let-shorthand'></a>(<a href='#prefer-if-let-shorthand'>link</a>) Omit the right-hand side of the expression when unwrapping an optional property to a non-optional property with the same name. [![SwiftFormat: redundantOptionalBinding](https://img.shields.io/badge/SwiftFormat-redundantOptionalBinding-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantOptionalBinding)
```

What do you think of making the link more descriptive, which is aligned with other links?

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -75,3 +76,4 @@
 --rules blankLinesAtEndOfScope
 --rules emptyBraces
 --rules andOperator
+\
```
bachand, line 79: typo?

README.md
```diff
@@ -739,12 +771,12 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   ```swift
   // WRONG
-  if let galaxy = galaxy,
+  if let galaxy,
```
bachand, line 774: Did you use a regex to search for other code examples to correct, or did you look through all of them?

bachand (approved):
Very excited about this one.

README.md
calda: I think I looked at all cases of `if let` and `guard let` in the document

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
calda: oops thanks!
```suggestion
```

PR #193, opened 2022-09-12
Add rule to prefer opaque generic parameter syntax

calda:
#### Summary

This PR proposes a rule to prefer opaque generic parameter syntax (using `some`) over the more-verbose named generic parameter syntax where equivalent.

```swift
// WRONG
func spaceshipDashboard<WarpDriveView: View, CaptainsLogView: View>(
  warpDrive: WarpDriveView,
  captainsLog: CaptainsLogView)
  -> some View
{ … }

func generate<Planets>(_ planets: Planets) where Planets: Collection, Planets.Element == Planet {
  …
}

// RIGHT
func spaceshipDashboard(
  warpDrive: some View,
  captainsLog: some View)
  -> some View
{ … }

func generate(_ planets: some Collection<Planet>) {
  …
}
```

#### Reasoning

We are adopting Swift 5.7 soon, so we should adopt guidance for its new language features.

Opaque generic parameter syntax is significantly less verbose and thus more legible than the full named generic parameter syntax. Auto-formatting for this rule is implemented in https://github.com/nicklockwood/SwiftFormat/pull/1206.

#### `some Any`

Fully-unconstrained generic parameters are somewhat uncommon, but are equivalent to `some Any`. For example:

```swift
func assertFailure<Value>(_ result: Result<Value, Error>) {
  if case .failure(let error) = result {
    XCTFail(error.localizedDescription)
  }
}

// is equivalent to:
func assertFailure(_ result: Result<some Any, Error>) {
  if case .failure(let error) = result {
    XCTFail(error.localizedDescription)
  }
}
```

`some Any` is somewhat unintuitive, and the named generic parameter is useful in this situation to compensate for the weak type information. Because of this, prefer using named generic parameters instead of `some Any`. 

With `--someAny disabled` (https://github.com/nicklockwood/SwiftFormat/pull/1261), the SwiftFormat rule never automatically converts generic functions to use `some Any`. This isn't outright banned, however, since we don't have an auto-formatting implementation that converts `some Any` _to_ named generic parameter syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2143,6 +2143,40 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='opaque-generic-parameters'></a>(<a href='#opaque-generic-parameters'>link</a>) Prefer using opaque generic parameters (with `some`) over verbose named generic parameter syntax where possible. [![SwiftFormat: opaqueGenericParameters](https://img.shields.io/badge/SwiftFormat-opaqueGenericParameters-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#opaqueGenericParameters)
+
+    <details>
+
+    #### Why?
+
+    Opaque generic parameter syntax is significantly less verbose and thus more legible than the full named generic parameter syntax.
+
+    ```swift
+    // WRONG
+    func spaceshipDashboard<WarpDriveView: View, CaptainsLogView: View>(
+      warpDrive: WarpDriveView,
+      captainsLogView: CaptainsLogView)
+      -> some View
+    { … }
+
+    func generate<Planets>(_ planets: Planets) where Planets: Collection, Planets.Element == Planet {
+      …
+    }
+
+    // RIGHT
+    func spaceshipDashboard(
+      warpDrive: some View,
+      captainsLogView: some View)
+      -> some View
+    { … }
+
+    func generate(_ planets: some Collection<Planet>) {
+      …
+    }
+    ```
+
+    </details>
+
 **[⬆ back to top](#table-of-contents)**
 
 ## File Organization
--- a/README.md
+++ b/README.md
@@ -2143,7 +2143,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
-* <a id='opaque-generic-parameters'></a>(<a href='#opaque-generic-parameters'>link</a>) **Prefer using opaque generic parameters (with `some`) over verbose named generic parameter syntax where possible.**  [![SwiftFormat: opaqueGenericParameters](https://img.shields.io/badge/SwiftFormat-opaqueGenericParameters-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#opaqueGenericParameters)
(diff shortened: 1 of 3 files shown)


erichoracek (approved):
Nice! Just one small suggestion

README.md
```diff
@@ -2143,6 +2143,61 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='opaque-generic-parameters'></a>(<a href='#opaque-generic-parameters'>link</a>) **Prefer using opaque generic parameters (with `some`) over verbose named generic parameter syntax where possible.**  [![SwiftFormat: opaqueGenericParameters](https://img.shields.io/badge/SwiftFormat-opaqueGenericParameters-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#opaqueGenericParameters)
+
+    <details>
+
+    #### Why?
+
+    Opaque generic parameter syntax is significantly less verbose and thus more legible than the full named generic parameter syntax.
```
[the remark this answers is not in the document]
calda: Good idea, I'll add a few examples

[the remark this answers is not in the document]
calda: https://github.com/airbnb/swift/pull/193/commits/b16ed9a9d4bd3a56d2967b95db5d5ad5f79dfd41

README.md
```diff
@@ -2143,6 +2143,72 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    ```
+
+    #### `some Any`
+
+    Fully-unconstrained generic parameters are somewhat uncommon, but are equivalent to `some Any`. For example:
+
+    ```swift
+    func assertFailure<Value>(_ result: Result<Value, Error>) {
+      if case .failure(let error) = result {
+        XCTFail(error.localizedDescription)
```
bachand, line 2196: My understanding is that we'd discourage writing an `XCTest` helper like this since then we would lose the file/line number of the failure (@fdiaz ). I wonder if we can come up with another example.

README.md
```diff
@@ -2143,6 +2143,72 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='opaque-generic-parameters'></a>(<a href='#opaque-generic-parameters'>link</a>) **Prefer using opaque generic parameters (with `some`) over verbose named generic parameter syntax where possible.**  [![SwiftFormat: opaqueGenericParameters](https://img.shields.io/badge/SwiftFormat-opaqueGenericParameters-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#opaqueGenericParameters)
```
bachand, line 2146: Link to proposal in case it's helpful: https://github.com/apple/swift-evolution/blob/main/proposals/0341-opaque-parameters.md

bachand (approved):

README.md
bachand, line 2146:
```suggestion
* <a id='prefer-opaque-generic-parameters'></a>(<a href='#prefer-opaque-generic-parameters'>link</a>) **Prefer using opaque generic parameters (with `some`) over verbose named generic parameter syntax where possible.**  [![SwiftFormat: opaqueGenericParameters](https://img.shields.io/badge/SwiftFormat-opaqueGenericParameters-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#opaqueGenericParameters)
```
What do you think of making the link more descriptive, which is aligned with many other rules.

README.md
calda: I pulled this example from the apps repo -- its sort of hard to come up with an example use case of a totally unconstrained generic parameter that doesn't feel contrived. Open to other suggestions though!

bachand: I agree that this is a good example so that I updated it to pass through the file and line: https://github.com/airbnb/swift/pull/204/

PR #198, opened 2022-09-30
Add rule to use doc comments before declarations, and regular comments elsewhere

calda:
#### Summary

This PR proposes a new rule to use doc comments (`///`) before declarations, and regular comments (`//`) elsewhere. 

This rule has autocomplete implemented by the new SwiftFormat `docComments` rule added in https://github.com/nicklockwood/SwiftFormat/pull/1266.

```swift
// WRONG

// A planet that exists somewhere in the universe.
class Planet {
  // Properties of a planet that are important for life
  let atmosphere: Atmosphere
  let oceans: [Ocean]

  // Terraforms the planet, by adding an atmosphere and ocean that is hospitable for life.
  func terraform() {
    /// This gas concentration has a pretty good track record so far
    let configuration = AtmosphereConfiguration(nitrogen: 0.78, oxygen: 0.22)
    
    /// Generate the atmosphere first, then the oceans
    generateAtmosphere(using: configuration)
    generateOceans()
  }
}

// RIGHT

class Planet {
  /// Properties of a planet that are important for life
  let atmosphere: Atmosphere
  let oceans: [Ocean]

  func terraform() {
    // This gas concentration has a pretty good track record so far
    let configuration = AtmosphereConfiguration(nitrogen: 0.78, oxygen: 0.22)
    
    // Generate the atmosphere first, then the oceans
    generateAtmosphere(using: configuration)
    generateOceans()
  }
}
```

#### Reasoning

Doc comments show up in the Xcode summary of a declaration, so we should prefer using doc comments for declarations. 

When not associated with a declaration, doc comments have no special meaning, so we should prefer regular comments elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1041,6 +1041,50 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
   ```
 
+* <a id='doc-comments'></a>(<a href='#doc-comments'>link</a>) Use doc comments (`///`) before declarations, and regular comments (`//`) elsewhere. [![SwiftFormat: docComments](https://img.shields.io/badge/SwiftFormat-docComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#docComments)
+
+  <details>
+
+  ```swift
+  // WRONG
+
+  // A planet that exists somewhere in the universe.
+  class Planet {
+    // Properties of a planet that are important for life
+    let atmosphere: Atmosphere
+    let oceans: [Ocean]
+
+    // Terraforms the planet, by adding an atmosphere and ocean that is hospitable for life.
+    func terraform() {
+      /// This gas concentration has a pretty good track record so far
+      let configuration = AtmosphereConfiguration(nitrogen: 0.78, oxygen: 0.22)
+      
+      /// Generate the atmosphere first, then the oceans
+      generateAtmosphere(using: configuration)
+      generateOceans()
+    }
+  }
+
+  // RIGHT
+
+  /// A planet that exists somewhere in the universe.
+  class Planet {
+    /// Properties of a planet that are important for life
+    let atmosphere: Atmosphere
+    let oceans: [Ocean]
+
+    /// Terraforms the planet, by adding an atmosphere and ocean that is hospitable for life.
+    func terraform() {
+      // This gas concentration has a pretty good track record so far
+      let configuration = AtmosphereConfiguration(nitrogen: 0.78, oxygen: 0.22)
+      
+      // Generate the atmosphere first, then the oceans
+      generateAtmosphere(using: configuration)
+      generateOceans()
+    }
+  }
+  ```
+
   </details>
 
 * <a id='whitespace-around-comment-delimiters'></a>(<a href='#whitespace-around-comment-delimiters'>link</a>) Include spaces or newlines before and after comment delimiters (`//`, `///`, `/*`, and `*/`) [![SwiftFormat: spaceAroundComments](https://img.shields.io/badge/SwiftFormat-spaceAroundComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#spaceAroundComments) [![SwiftFormat: spaceInsideComments](https://img.shields.io/badge/SwiftFormat-spaceInsideComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#spaceInsideComments)
--- a/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
+++ b/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
@@ -69,6 +69,7 @@
 --rules spaceAroundParens
 --rules enumNamespaces
 --rules blockComments
+--rules docComments
 --rules spaceAroundComments
 --rules spaceInsideComments
 --rules blankLinesAtStartOfScope


Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -69,6 +69,7 @@
 --rules spaceAroundParens
 --rules enumNamespaces
 --rules blockComments
+--rules docComments
```
calda, line 72: This will fail in CI until we pull in a version of SwiftFormat that includes this rule implementation

README.md
```diff
@@ -1041,6 +1041,50 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
 
+* <a id='doc-comments'></a>(<a href='#doc-comments'>link</a>) Use doc comments (`///`) before declarations, and regular comments (`//`) elsewhere. [![SwiftFormat: docComments](https://img.shields.io/badge/SwiftFormat-docComments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#docComments)
+
+  <details>
+
+  ```swift
+  // WRONG
+
+  // A planet that exists somewhere in the universe.
+  class Planet {
```
bachand, line 1052:
```suggestion
  final class Planet {
```

README.md
```diff
@@ -1041,6 +1041,50 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+      /// Generate the atmosphere first, then the oceans
+      generateAtmosphere(using: configuration)
+      generateOceans()
+    }
+  }
+
+  // RIGHT
+
+  /// A planet that exists somewhere in the universe.
+  class Planet {
```
bachand, line 1071:
```suggestion
  final class Planet {
```

README.md
bachand: Since we say to prefer `final` this seems like the right way to write this example?

bachand (approved):

swiftal64:
@calda There are several reasons I suggest we don't implement this rule:

**1. Documenting local variables is not allowed anymore**

    /// This gas concentration has a pretty good track record so far
    let configuration = AtmosphereConfiguration(nitrogen: 0.78, oxygen: 0.22)

... is actually better since option-clicking on `configuration` gives us the description of the variable.

**2. Markdown/jumping in non-doc comments not allowed anymore**

Occasionally I'll leave a long block comment describing what's going on in a function. Although it doesn't show up in documentation, the main advantage of `///` is markdown support (including clickable links). Hypothetical example:

```
func calculateLargestPlanetMass() -> Float {
  let largestPlanet = planets.max()

  /// We **must** multiply by factor here because:
  ///
  /// 1. There is a legacy bug that consumers are expecting the factor to be included.
  /// 2. When we enable scaling in the future, the `factor` variable will be used for that.
  ///
  /// See ``Environment`` for more information.
  return largestPlanet.mass * factor
}
```

Additionally, Xcode allows me to command+click on `Environment` to jump to its definition. It does not allow me to do so if it is in a `//` comment. Even though we never see this documentation inside Xcode UI, the Xcode editor will convert identifiers like `must` into bold.

**3. Might mess up [compiler directives](https://useyourloaf.com/blog/todo-fixme-and-compiler-directives/)**

Occasionally we have both a `TODO` and a doc comment at the top, I suspect this rule will remove the properly formatted `TODO` (which traditionally has `//` in front):

```
// TODO: Rename to AnyPlanet
class Planet {
}
```

Similarly, `MARK` must be preceded by `//` to function properly within the Xcode UI, so we should make sure we aren't accidentally replacing those:

```
// MARK: Environment classes
class Planet {
}
```

There are other keywords like these (e.g. `FIXME`).

**4. Doc about individual definition vs. group of definitions**

In the following example, we assume that `Internal types` applies to both `unichar` and `keyValue`:

```
/// Labels
enum JSONLabel: String {
    case null
    case boolean
    case number
    case string
    case json
    case array
    case object

    // Internal types
    case unichar
    case keyValue
}
```

If we convert it to a `///`, we now see "Internal types" when we option-click `unichar` but not when we option-click `keyValue`. I think we should allow both `///` and `//` inside enum cases with the former being used to describe one element, and the latter being used to describe all the following elements. Of course the best solution would be to expand it, but this is verbose and I'm sure a lot of places exist where we would need to manually fix this:

```
/// Labels
enum JSONLabel: String {
    case null
    case boolean
    case number
    case string
    case json
    case array
    case object

    /// Internal type
    case unichar
    /// Internal type
    case keyValue
}
```

-

Similarly, in this other example, `//` should be preferred over `///` since it's describing all 3 lines:

```
// Prevent control characters confusing parser
private let start = UnicodeScalar(63232)!
private let end = UnicodeScalar(63235)!
private let cursorCharacters = CharacterSet(charactersIn: start ... end)
```

**5. External vs. internal documentation no longer allowed**

`///` is inherently used for external documentation (i.e. for consumers) since you can option-click the identifier and read its documentation. `//` therefore is primarily used for internal documentation. Occasionally I may want to document implementation details which the consumer should not care about. In those cases, I prefer to use `//` since otherwise it adds noise to consumers. Here's an example:

```
struct PlanetVendor {
  // This is a function (`() -> AnyPlanet`) instead of variable (`AnyPlanet`) because the
  // Universe class does not allow us to show a planet once it's been dismissed.
  /// Construct your planet in this function.
  var makePlanet: () -> AnyPlanet
}
```

---

It's going to be very hard to know when someone intentionally chose to use the wrong comment type; e.g. `///` for markdown purposes on a type, `//` to document a group of lines, etc. 

Therefore, I don't think this broad rule should be enabled. If we could get a more constrained version that has less false positives (e.g. only add `///` comments to types that likely are never grouped like `class` or `struct`), that could be one direction it could go, but even then it might be so constrained and have many edge cases (e.g. `// TODO` comments must come before the `///` comments so as not to break Xcode documentation) that it might not be worth it.

calda:
Thanks for the feedback @swiftal64, I think you're right that this is too nuanced to autocorrect.

PR #204, opened 2022-10-21
Update `prefer-opaque-generic-parameters` XCTest example to pass through `file` and `line` 

bachand:
#### Summary

It is a best practice when writing assertion helpers to pass through the file and line from the caller's context. This ensures that errors surfaced by XCTest are most helpful. Here I update the code example in our style guide to follow that practice.

#### Reasoning

We strive to have our code examples follow known best practices.

@airbnb/swift-styleguide-maintainers


calda (approved):
Thanks!

PR #206, opened 2022-11-04
Update `Package.swift` to explicitly not support Linux

calda:
Our Swift Package Index badge in the README currently says this package supports Linux:

image

The package builds correctly on Linux, which is what SPI currently uses to detect platform compatibility. The plugin fails at runtime on Linux, though, because we have binary dependencies that only support macOS: https://github.com/airbnb/swift/discussions/197

The maintainer of Swift Package Index suggested this workaround (https://github.com/airbnb/swift/discussions/197#discussioncomment-4055303).


calda:
We could alternatively add support for Linux by building our binary dependencies with Linux support, but I rebuild the SwiftFormat dependency manually somewhat often and I'm a bit concerned about the additional overhead.

bachand (approved):

PR #212, opened 2023-02-09
Fixed typos in README.md

teddy5518:
#### Summary
Fixed typos

#### Reasoning
for better doc clarity

cc. @calda

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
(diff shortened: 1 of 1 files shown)


calda:
These don't seem to be typos -- I think we can keeping the grammar / formatting as-is.

teddy5518:
@calda Thank you for a quick review. :)

Can we discuss about the proposal?

I believe first change is needed as the whole sentence is not quoted, but the quotation mark was used to emphasize part of sentence. Thus, I believe adjusting placement of punctuation increases readability of the doc.

The second change was more of recommendation as following example was using similar but different style of wording. I thought it would be better to align style of expression.

PR #214, opened 2023-03-20
Clarify Rule: Never use the default case when switching over an enum

teddy5518:
#### Summary
This PR proposes clarifying the rule of "Never use the default case when switching over an enum" to be more specific.

#### Reasoning
I clarified rule with "real world" example and to follow clear and understandable examples that the guide is providing. (such as universe, weather and etc.)

Please review. Thanks! 😄

Reviewers
cc. @calda

### DIFF
--- a/README.md
+++ b/README.md
@@ -2027,19 +2027,19 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   ```swift
   // WRONG
-  switch anEnum {
-  case .a:
-    // Do something
+  switch trafficLight {
+  case .greenLight:
+    // Move your vehicle
   default:
-    // Do something else.
+    // Stop your vehicle
   }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2027,19 +2027,19 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  switch trafficLight {
+  case .greenLight:
+    // Move your vehicle
   default:
-    // Do something else.
+    // Stop your vehicle
   }
 
   // RIGHT
   switch anEnum {
```
calda, line 2038: Should this be `switch trafficLight {`?

calda (approved):
Nice improvement!

bachand (approved):
Love this. Thanks!

PR #215, opened 2023-03-21
Clarify Rule: Use Swift's automatic enum values unless they map to an external source.

teddy5518:
#### Summary

This PR proposes to clarify comment of the "Use Swift's automatic enum values unless they map to an external source." 

#### Reasoning

1) I indicated whether the examples are wrong or right like other rules in the guide to increase readability.
2) The origin comment was not being specific as it only said "swiftformat:enable redundantRawValues" and "swiftformat:disable redundantRawValues". The heading of rule provides link to swiftformat, so I thought it was better to directly indicate that the examples are right for relying on Swift's automatic enum values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1854,6 +1854,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   // RIGHT
   // Relying on Swift's automatic enum values
+  // swiftformat:disable redundantRawValues
   enum Planet: Int {
     case mercury
     case venus
@@ -1864,6 +1865,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
     case uranus
     case neptune
   }
+  // swiftformat:enable redundantRawValues
 
   // RIGHT
   /// These values come from the server, so we set them here explicitly to match those values.
--- a/README.md
+++ b/README.md
@@ -1809,12 +1809,14 @@ _You can enable the following settings in Xcode by running [this script](resourc
     case warning = "warning"
   }
 
+  // WRONG
   enum UserType: String {
     case owner
     case manager
     case member
   }
 
+  // WRONG
   enum Planet: Int {
     case mercury = 0
     case venus = 1
@@ -1826,6 +1828,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
     case neptune = 7
   }
 
+  // WRONG
   enum ErrorCode: Int {
     case notEnoughMemory
     case invalidResource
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1826,27 +1828,30 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
   // RIGHT
+  // Relying on Swift's automatic enum values
   enum ErrorType: String {
     case error
     case warning
   }
 
+  // RIGHT
   /// These are written to a logging service. Explicit values ensure they're consistent across binaries.
-  // swiftformat:disable redundantRawValues
```
calda, line 1842: We should keep this example linter directive, since otherwise the linter would remove the redundant raw values

teddy5518: @calda Thank you for your explanation! I learned something new today! I have fixed accordingly. Please review again. 😄

README.md
```diff
@@ -1826,27 +1828,32 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  /// These are written to a logging service. Explicit values ensure they're consistent across binaries.
   enum UserType: String {
     case owner = "owner"
     case manager = "manager"
     case member = "member"
   }
-  // swiftformat:enable redundantRawValues
 
+  // RIGHT
+  // swiftformat:enable redundantRawValues
```
calda, line 1855: This should be:

```suggestion
  // swiftformat:disable redundantRawValues
```

with the corresponding `// swiftformat:enable redundantRawValues` directive after the closing `}` of the enum

README.md
```diff
@@ -1826,27 +1828,32 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
-  /// These are written to a logging service. Explicit values ensure they're consistent across binaries.
+  // RIGHT
   // swiftformat:disable redundantRawValues
+  /// These are written to a logging service. Explicit values ensure they're consistent across binaries.
   enum UserType: String {
     case owner = "owner"
     case manager = "manager"
     case member = "member"
   }
-  // swiftformat:enable redundantRawValues
```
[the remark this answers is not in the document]
teddy5518: @calda Now, I totally understand why `swiftformat:disable redundantRawValues` and `swiftformat:enable redundantRawValues`were there. Thank you so much for your explanation. I truly appreciate your time and effort.

README.md
teddy5518: @calda Sorry to bother you again, but I am confused with "redundant" raw values and usage of swiftformat. 

As I have read rules.md of swiftformat and your comment for serveral times, I think 
1) `enum ErrorType: String' in WRONG is the one that needs to be implemented with `swiftformat:disable redundantRawValues` and `swiftformat:enable redundantRawValues` as it needs to be excepted from swiftformat rules.
2) `enum Planet: Int` needs to be reverted as before, becuase it is straightly following Swift's automatic enum values, and it does not contain any redundant raw values.

May I hear back what you think on these two?
Thank you for your time in advance.

calda: Oops! Sorry, you're right -- my mistake. Mind making another PR to fix that?

teddy5518: Thank you for getting back to me! I am on it. Allow me few seconds please! :)

PR #216, opened 2023-03-21
Clarify Rule: Use Swift's automatic enum values unless they map to an external source.

teddy5518:
#### Summary

This PR fixes placement of `swiftformat:disable redundantRawValues` and `swiftformat:enable redundantRawValues`.

#### Reasoning

[As discussed here](https://github.com/airbnb/swift/pull/215#discussion_r1143742872), enum `ErrorType: String'  in WRONG is the one that needs to be implemented with `swiftformat:disable redundantRawValues` and `swiftformat:enable redundantRawValues` as it needs to be excepted from swiftformat rules.
And `enum Planet: Int` needs to be reverted as before, because it is straightly following Swift's automatic enum values, and it does not contain any redundant raw values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1804,10 +1804,12 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   ```swift
   // WRONG
+  // swiftformat:disable redundantRawValues
   enum ErrorType: String {
     case error = "error"
     case warning = "warning"
   }
+  // swiftformat:enable redundantRawValues
 
   // WRONG
   enum UserType: String {
@@ -1854,7 +1856,6 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   // RIGHT
   // Relying on Swift's automatic enum values
-  // swiftformat:disable redundantRawValues
   enum Planet: Int {
     case mercury
     case venus
@@ -1865,7 +1866,6 @@ _You can enable the following settings in Xcode by running [this script](resourc
     case uranus
     case neptune
   }
-  // swiftformat:enable redundantRawValues
 
   // RIGHT
   /// These values come from the server, so we set them here explicitly to match those values.


README.md
```diff
@@ -1804,10 +1804,12 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   ```swift
   // WRONG
+  // swiftformat:disable redundantRawValues
```
calda, line 1807: Why is this "wrong"?

README.md
```diff
@@ -1854,7 +1856,6 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   // RIGHT
   // Relying on Swift's automatic enum values
-  // swiftformat:disable redundantRawValues
```
calda, line 1857: Thanks 👍🏻

README.md
teddy5518: @calda Are we talking "wrong" as comment "// WRONG"?

calda: I'm not sure I understand why this needs both `// WRONG` _and_ `// swiftformat:disable redundantRawValues`. Doesn't the SwiftFormat directive mean this is correct (e.g. we're manually indicating that these values should be preserved because they map to an external source)?

teddy5518: It was there even before I requested PR #216 and PR #215 . And, I thought it was there as it was not relying on Swift's automatic enum values. (Error could be coming from internal source.)

If you think otherwise then shouldn't line 1807 to 1812 (enum ErrorType) switched with 1841 to 1845 (enum ErrorType)?

teddy5518: > I'm not sure I understand why this needs both `// WRONG` _and_ `// swiftformat:disable redundantRawValues`. Doesn't the SwiftFormat directive mean this is correct (e.g. we're manually indicating that these values should be preserved because they map to an external source)?

Oh, I thought we were agreed to state it as wrong and make exception to show example of wrongdoings.

calda: haha yeah I think removing the directive is fine, otherwise this may be confusing

calda (approved):

PR #219, opened 2023-03-31
Update: ignore Rule OpenSource Library

interactord:
#### Summary

I have added code to two source files excluding external libraries such as Carthage, Pods, and SourcePackages (SPM).

#### Reasoning

I added code to enforce the Airbnb style rules only on the code we wrote, as I believe it is the intended behavior to apply the rules only to our own code when using the command "swiftFormat .", which also modifies external libraries.

### DIFF
--- a/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
+++ b/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
@@ -1,5 +1,5 @@
 # Exclude checkout directories for common package managers
---exclude Carthage,Pods
+--exclude Carthage,Pods,.build
 
 # options
 --swiftversion 5.7
--- a/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
+++ b/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
@@ -1,3 +1,6 @@
+# Exclude checkout directories for common package managers
+--exclude Carthage,Pods
+
 # options
 --swiftversion 5.7
 --self remove # redundantSelf
--- a/Sources/AirbnbSwiftFormatTool/swiftlint.yml
+++ b/Sources/AirbnbSwiftFormatTool/swiftlint.yml
@@ -17,6 +17,7 @@ only_rules:
 excluded:
   - Carthage
   - Pods
+  - .build
 
 colon:
   apply_to_dictionaries: false


Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -1,3 +1,6 @@
+# ignore
+--exclude Carthage,Pods,SourcePackages
```
calda, line 2: What is SourcePackages?

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -1,3 +1,6 @@
+# ignore
```
calda, line 1:
```suggestion
# Exclude checkout directories for common package managers
```

interactord: I like your comment. It's clearer than mine.

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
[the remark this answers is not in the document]
interactord: The two files, swiftLint and swiftFormatter, do not run simultaneously as each executable operates differently.

Therefore, I believe it is better to exclude both files clearly, so I added them to the exclusion list.

calda: In a Swift project / git repo, where is this directory? Is it at the repo root, or is it somewhere else?

interactord: It is located at the root of the repository.

interactord:
In the previous PR, I attempted to remove the SourcePackages file from the lint. However, upon reviewing the FastFile code again, I realized that SourcePackages was a custom name used in my project for caching SPM libraries in fastlane.

Therefore, I decided that this was not a part to request changes for in the PR and made the necessary modifications before committing again. Additionally, I believe it would be a good idea to explicitly specify external libraries not only in SwiftLint but also in swiftFormat.

Furthermore, the .build folder is generated by the swift package build command and does not require linting, so I made additional modifications to update the source accordingly.

calda (approved):
Thanks!

PR #227, opened 2023-05-08
Add rule that protocol composition type aliases should be sorted alphabetically

calda:
#### Summary

This PR proposes a new rule that protocol composition type aliases should be sorted alphabetically.

Autocorrect is implemented in the SwiftFormat `sortTypealiases` rule, added in https://github.com/nicklockwood/SwiftFormat/pull/1445.

```swift
// WRONG (not sorted)
public typealias Dependencies
  = UniverseBuilderProviding
  & LawsOfPhysicsProviding
  & UniverseSimulatorServiceProviding
  & PlanetBuilderProviding
  & CivilizationServiceProviding

// RIGHT
public typealias Dependencies
  = CivilizationServiceProviding
  & LawsOfPhysicsProviding
  & PlanetBuilderProviding
  & UniverseBuilderProviding
  & UniverseSimulatorServiceProviding
```

#### Reasoning

Type aliases are an unordered list with no natural ordering. Sorting alphabetically keeps these lists more organized, which is especially valuable for long protocol compositions.

_Please react with 👍/👎 if you agree or disagree with this proposal._

PR #228, opened 2023-05-08
Add rule to add a linebreak before the else keyword in a multi-line guard statement

calda:
#### Summary

This PR proposes a new rule to add a linebreak before the else keyword in a multi-line guard statement.

```swift
// WRONG -- else should be on its own line for multi-line guard statements
guard
  let galaxy,
  galaxy.name == "Milky Way" else
{ … }

// WRONG -- else should be on the same line for single-line guard statements
guard let galaxy
else { … }

// RIGHT
guard
  let galaxy,
  galaxy.name == "Milky Way"
else { … }

// RIGHT
guard let galaxy else {
  …
}
```

#### Reasoning

This was first suggested back in https://github.com/airbnb/swift/pull/108#issuecomment-715479492. This is also how the examples in the "wrap multi-line conditionals" rule description are formatted.

Placing the `else` on its own line makes it more visually distinct, and prevents it from blending in with the last statement.

We previously didn't enforce this through the formatter since the SwiftFormat rule for this didn't match our requirements very well. As of https://github.com/nicklockwood/SwiftFormat/pull/1447 the SwiftFormat rule implementation works well for us.

### DIFF
--- a/README.md
+++ b/README.md
@@ -843,18 +843,18 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
   ```
 
-* <a id='wrap-guard-else'></a>(<a href='#wrap-guard-else'>link</a>) **Break before the `else` keyword in a multi-line guard statement.** [![SwiftFormat: elseOnSameLine](https://img.shields.io/badge/SwiftFormat-elseOnSameLine-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#elseOnSameLine )
+* <a id='wrap-guard-else'></a>(<a href='#wrap-guard-else'>link</a>) **Add a line break before the `else` keyword in a multi-line guard statement.** For single-line guard statements, keep the `else` keyword on the same line as the `guard` keyword. The open brace should immediately follow the `else` keyword. [![SwiftFormat: elseOnSameLine](https://img.shields.io/badge/SwiftFormat-elseOnSameLine-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#elseOnSameLine )
 
   <details>
 
   ```swift
-  // WRONG -- else should be on its own line for multi-line guard statements
+  // WRONG (else should be on its own line for multi-line guard statements)
   guard
     let galaxy,
     galaxy.name == "Milky Way" else
   { … }
 
-  // WRONG -- else should be on the same line for single-line guard statements
+  // WRONG (else should be on the same line for single-line guard statements)
   guard let galaxy
   else { … }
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -817,6 +817,33 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
   ```
 
+* <a id='wrap-guard-else'></a>(<a href='#wrap-guard-else'>link</a>) **Break before the `else` keyword in a multi-line guard statement.** [![SwiftFormat: elseOnSameLine](https://img.shields.io/badge/SwiftFormat-elseOnSameLine-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#elseOnSameLine )
+
+  <details>
+
+  ```swift
+  // WRONG -- else should be on its own line for multi-line guard statements
```
erichoracek, line 825: In https://github.com/airbnb/swift/pull/227 we use parens for explaining why this is wrong, should we use that here too?
```suggestion
  // WRONG (else should be on its own line for multi-line guard statements)
```

README.md
```diff
@@ -817,6 +817,33 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  guard
+    let galaxy,
+    galaxy.name == "Milky Way" else
+  { … }
+
+  // WRONG -- else should be on the same line for single-line guard statements
```
erichoracek, line 831:
```suggestion
  // WRONG (else should be on the same line for single-line guard statements)
```

README.md
```diff
@@ -817,6 +817,33 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
   ```
 
+* <a id='wrap-guard-else'></a>(<a href='#wrap-guard-else'>link</a>) **Break before the `else` keyword in a multi-line guard statement.** [![SwiftFormat: elseOnSameLine](https://img.shields.io/badge/SwiftFormat-elseOnSameLine-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#elseOnSameLine )
```
erichoracek, line 820: I don't think we're capturing that you shouldn't break before the `else` in a single line guard in this description?

erichoracek, line 820:
```suggestion
* <a id='wrap-guard-else'></a>(<a href='#wrap-guard-else'>link</a>) **Add a line break before the `else` keyword in a multi-line guard statement.** [![SwiftFormat: elseOnSameLine](https://img.shields.io/badge/SwiftFormat-elseOnSameLine-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#elseOnSameLine )
```

erichoracek (approved):

README.md
calda: Looks like in other example like this we also use parenthesis, so I'll update this to do that instead

PR #231, opened 2023-06-15
Add rules to discourage #file and #filePath

jqsilver:
#### Summary

Add lint rules to discourage the use of `#file` and `#filePath` literals (macros in 5.9). 

In addition, my editor removed some trailing whitespace - let me know if I should revert that. 

#### Reasoning

1. `#file`'s behavior is changing in future Swift versions (https://developer.apple.com/documentation/swift/file), from the problematic behavior of `#filePath` to the more ideal behavior of `#fileID`. We should just use the literal/macro with explicit and unchanging behavior. 
2. The documentation for `#filePath` says "Because #fileID doesn’t embed the full path to the source file, unlike #filePath, it gives you better privacy and reduces the size of the compiled binary. Avoid using #filePath outside of tests, build scripts, or other code that doesn’t become part of the shipping program." https://developer.apple.com/documentation/swift/filepath, so we should reflect this guidance in our lint rules. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -2323,7 +2323,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='no-filepath-literal'></a>(<a href='#no-file-literal'>link</a>) **Don't use `#filePath` in production code. Use `#fileID`.**
+* <a id='no-filepath-literal'></a>(<a href='#no-file-literal'>link</a>) **Don't use `#filePath` in production code. Use `#fileID` instead.**
 
   <details>
 
--- a/README.md
+++ b/README.md
@@ -2315,7 +2315,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   <details>
 
   #### Why?
-  The behavior of the `#file` literal (or macro as of Swift 5.9) has evolved from evaluating to the full source file path (the behavior as `#filePath`) to a human-readable string containing module and file name (the behavior of `#fileID`). Use the literal (or macro) with the more explicit behavior.
+  The behavior of the `#file` literal (or macro as of Swift 5.9) has evolved from evaluating to the full source file path (the behavior as of filePath`) to a human-readable string containing module and file name (the behavior of `#fileID`). Use the literal (or macro) with the most appropriate behavior for your use case.
 
   [Swift documentation](https://developer.apple.com/documentation/swift/file)
 
--- a/README.md
+++ b/README.md
@@ -2315,7 +2315,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   <details>
 
   #### Why?
-  The behavior of the #file literal (or macro as of Swift 5.9) has evolved from evaluating to the full source file path (the behavior as #filePath) to a human-readable string containing module and file name (the behavior of #fileID). Use the literal (or macro) with the more explicit behavior.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2310,14 +2310,38 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  #### Why?
+  The behavior of the `#file` literal (or macro as of Swift 5.9) has evolved from evaluating to the full source file path (the behavior as `#filePath`) to a human-readable string containing module and file name (the behavior of `#fileID`). Use the literal (or macro) with the more explicit behavior.
+
+  [Swift documentation](https://developer.apple.com/documentation/swift/file)
+
+  [Swift Evolution Proposal: Concise magic file names](https://github.com/apple/swift-evolution/blob/main/proposals/0274-magic-file.md)
+
+  </details>
+
+* <a id='no-filepath-literal'></a>(<a href='#no-file-literal'>link</a>) **Don't use `#filePath` in production code. Use `#fileID`.**
```
calda, line 2326:
```suggestion
* <a id='no-filepath-literal'></a>(<a href='#no-file-literal'>link</a>) **Don't use `#filePath` in production code. Use `#fileID` instead.**
```

README.md
```diff
@@ -2310,14 +2310,38 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  <details>
+
+  #### Why?
+  `#filePath` should only be used in non-production code where the full path of the source file provides useful information to developers. Because `#fileID` doesn’t embed the full path to the source file, so it won't expose your file system and reduces the size of the compiled binary.
```
calda, line 2331:
```suggestion
  `#filePath` should only be used in non-production code where the full path of the source file provides useful information to developers. Because `#fileID` doesn’t embed the full path to the source file, it won't expose your file system and reduces the size of the compiled binary.
```

calda (approved):

calda:
@jqsilver, could you also add a lint rule for this to https://github.com/airbnb/swift/blob/master/Sources/AirbnbSwiftFormatTool/swiftlint.yml?

README.md
```diff
@@ -2310,14 +2310,38 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
+* <a id='no-file-literal'></a>(<a href='#no-file-literal'>link</a>) **Don't use #file. Use #fileID or #filePath as appropriate.**
+
+  <details>
+
+  #### Why?
+  The behavior of the `#file` literal (or macro as of Swift 5.9) has evolved from evaluating to the full source file path (the behavior as `#filePath`) to a human-readable string containing module and file name (the behavior of `#fileID`). Use the literal (or macro) with the more explicit behavior.
```
calda, line 2318:
```suggestion
  The behavior of the `#file` literal (or macro as of Swift 5.9) has evolved from evaluating to the full source file path (the behavior as of filePath`) to a human-readable string containing module and file name (the behavior of `#fileID`). Use the literal (or macro) with the more explicit behavior.
```

calda:
We can follow how the `no_direct_standard_out_logs` rule is set up, and not exempt test targets in the `swiftlint.yml` file in this repo (even though we exempt test targets / demo modules etc in the `swiftlint.yml` file in the apps repo).

PR #235, opened 2023-09-05
Include version-specific package manifests when determining default minimum Swift version

calda:
The Airbnb Swift Format Tool currently infers a package's minimum Swift supported version by taking the `// swift-tools-version` from the `Package.swift` (e.g. `// swift-tools-version:5.6`).

A project can also have multiple `Package.swift`s with different tools versions. This is required if a package wants to make use of functionality that is only available in later tools versions, without dropping support for earlier tool versions. For example, you can use `// swift-tools-version:5.8` in your `Package.swift` and have an additional `Package@swift-5.7.swift` with `// swift-tools-version:5.7`.

This PR updates the AirbnbSwift Format Tool to take into account all of these versions, and use the lowest value as the minimum Swift version when running the code formatter.

### DIFF
--- a/Plugins/FormatSwift/Plugin.swift
+++ b/Plugins/FormatSwift/Plugin.swift
@@ -92,9 +92,9 @@ extension AirbnbSwiftFormatPlugin: CommandPlugin {
       inputPaths = try self.inputPaths(for: context.package)
     }
 
-    // When running on a SPM package we infer the minimum Swift version from the
-    // `swift-tools-version` in `Package.swift` by default if the user doesn't
-    // specify one manually
+    // When running on a SPM package, by default we use the minimum swift version
+    // specified in any of the package manifest files. Alternatively the user can
+    // manually specify a swift version via the `--swift-version` argument.
     lazy var minimumSwiftVersion = context.package.minimumSwiftVersion
     let swiftVersion = argumentExtractor.extractOption(named: "swift-version").last
       ?? "\(minimumSwiftVersion.major).\(minimumSwiftVersion.minor)"
@@ -211,29 +211,20 @@ extension Package {
     ]
 
     // Look for all of the package manifest files in the directory root
-    let enumerator = FileManager.default.enumerator(
+    let filesInRootDirectory = try? FileManager.default.contentsOfDirectory(
       at: projectDirectory,
-      includingPropertiesForKeys: nil,
-      options: [.skipsSubdirectoryDescendants, .skipsHiddenFiles])
+      includingPropertiesForKeys: nil)
 
-    while let fileURL = enumerator?.nextObject() as? URL {
+    for fileURL in filesInRootDirectory ?? [] {
       let fileName = fileURL.lastPathComponent
 
       guard fileName.hasPrefix("Package"), fileName.hasSuffix(".swift") else { continue }
 
-      // Parse the Swift version from the file name if it has a format like `Package@swift-5.8.swift`
-      if
-        fileName.hasPrefix("Package@swift-"),
-        let swiftVersion = parseSwiftVersion(from: fileName.dropFirst("Package@swift-".count))
-      {
-        supportedSwiftVersions.append(swiftVersion)
-      }
-
       // Parse the Swift tools version from the file body if it starts with a comment like `// swift-tools-version: 5.8`
       if
         let fileContents = try? String(contentsOf: fileURL),
-        fileContents.hasPrefix("// swift-tools-version: "),
-        let swiftVersion = parseSwiftVersion(from: fileContents.dropFirst("// swift-tools-version: ".count))
+        fileContents.hasPrefix("// swift-tools-version:"),
+        let swiftVersion = parseSwiftVersion(from: fileContents.dropFirst("// swift-tools-version:".count))
       {
         supportedSwiftVersions.append(swiftVersion)
       }
@@ -244,13 +235,13 @@ extension Package {
 
   /// Parses a Swift version from a string like "5.8"
   private func parseSwiftVersion(from string: Substring) -> SwiftVersion? {
-    var string = string
+    var string = Substring(string.trimmingCharacters(in: .whitespacesAndNewlines))
 
     guard
       string.count >= 3,
-      let major = string.popFirst().flatMap({ Int("\($0)") }),
+      let major = string.popFirst().flatMap({ Int(String($0)) }),
       string.popFirst() == ".",
-      let minor = string.popFirst().flatMap({ Int("\($0)") })
+      let minor = string.popFirst().flatMap({ Int(String($0)) })
     else { return nil }
 
     return SwiftVersion(major: major, minor: minor)
--- a/Plugins/FormatSwift/Plugin.swift
+++ b/Plugins/FormatSwift/Plugin.swift
@@ -95,8 +95,9 @@ extension AirbnbSwiftFormatPlugin: CommandPlugin {
     // When running on a SPM package we infer the minimum Swift version from the
     // `swift-tools-version` in `Package.swift` by default if the user doesn't
     // specify one manually
+    lazy var minimumSwiftVersion = context.package.minimumSwiftVersion
     let swiftVersion = argumentExtractor.extractOption(named: "swift-version").last
-      ?? "\(context.package.toolsVersion.major).\(context.package.toolsVersion.minor)"
+      ?? "\(minimumSwiftVersion.major).\(minimumSwiftVersion.minor)"
 
     let arguments = [
       "--swift-version",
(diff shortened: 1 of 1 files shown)


Plugins/FormatSwift/Plugin.swift
```diff
@@ -95,8 +95,9 @@ extension AirbnbSwiftFormatPlugin: CommandPlugin {
     // When running on a SPM package we infer the minimum Swift version from the
     // `swift-tools-version` in `Package.swift` by default if the user doesn't
     // specify one manually
```
bachand, line 97: Putting here so we have a thread...

>A project can also have multiple Package.swifts with different tools versions. This is required if a package wants to make use of functionality that is only available in later tools versions, without dropping support for later tool versions. 

For the last sentence, do you mean this? (change in italics)

> This is required if a package wants to make use of functionality that is only available in later tools versions, without dropping support for _earlier_ tool versions. 

I want to make sure that I'm following.

Plugins/FormatSwift/Plugin.swift
```diff
@@ -166,3 +167,92 @@ enum CommandError: Error {
...
+  var minimumSwiftVersion: SwiftVersion {
+    supportedSwiftVersions.sorted().first!
+  }
+
+  // MARK: Private
+
+  /// Swift versions supported by this package. Guaranteed to be non-empty.
+  ///  - This includes the `swift-tools-version` from the `Package.swift`,
+  ///    plus the Swift version of any additional version-specific Package manifest
+  ///    (e.g. `Package@swift-5.6.swift`).
```
bachand, line 205: 👍

Plugins/FormatSwift/Plugin.swift
```diff
@@ -166,3 +167,92 @@ enum CommandError: Error {
...
+  private var supportedSwiftVersions: [SwiftVersion] {
+    guard let projectDirectory = URL(string: directory.string) else { return [] }
```
bachand, line 207: @calda can you help me follow where the `directory` property is defined? I didn't see it listed here: https://developer.apple.com/documentation/packagedescription/package.

Plugins/FormatSwift/Plugin.swift
```diff
@@ -166,3 +167,92 @@ enum CommandError: Error {
...
+    }
+  }
+}
+
+extension Package {
+
+  // MARK: Internal
+
+  /// The minimum Swift version supported by this package
+  var minimumSwiftVersion: SwiftVersion {
```
bachand, line 196: How feasible would it be to add a couple of unit tests for this method?

bachand (commented):
@calda I love that we're going to support more complex setups with multiple Package.swift files!

I am reviewing this PR before completing my review for https://github.com/airbnb/swift/pull/236 since I see that https://github.com/airbnb/swift/pull/236 depends on this PR.

I have a few questions I wanted to ask before continuing my review on this PR 👍

Plugins/FormatSwift/Plugin.swift
calda: Ah yes thank you, that was a typo.

Plugins/FormatSwift/Plugin.swift
calda: I think this is the `PackagePlugin.Package` type, which is different from `PackageDescription.Package`. I can't actually find documentation online for the `PackagePlugin` framework, but it exposes the APIs that you use in SPM package plugins.

Plugins/FormatSwift/Plugin.swift
calda: We don't currently have any unit tests for this package. What we do have is a CI job that builds and runs the package plugin, so verifies that any functionality exercised in the command builds and runs successfully.

In our case, as of https://github.com/airbnb/swift/pull/236 CI will fail if this doesn't work correctly. That provides a minimum guarantee that this works as expected.

Unit testing this logic would be a bit involved, especially since it works with the file system. We'd have to restructure the code to not directly interact with the filesystem, which adds complexity and overhead.

I find the existence of a CI job that fails without this to be a sufficient test case. Thoughts?

bachand (approved):
Nice work @calda !

Plugins/FormatSwift/Plugin.swift
calda: I looked in to testing this logic, and there are some obstacles:

First, it isn't possible to unit test logic in a Swift package plugin. There is confirmation about this here: https://forums.swift.org/t/how-can-we-test-swiftpm-plugins/61279/2

After that didn't work I figured we can still extract the code we wanted to test into a new, separate module called "PluginHelpers" which can then be imported by the plugin. Unfortunately package plugins also can't even import other modules -- I get this error:

```
plugin 'FormatSwift' cannot depend on 'PluginHelpers' of type 'library'; this dependency is unsupported
```

If we want to test this logic we'd have to move it out of the plugin entirely and into the `AirbnbSwiftFormatTool` executable. That would add additional complexity to our communication between the plugin and executable, since we have to create a new way to represent all of the needed info in the arguments we use when invoking the executable.

Plugins/FormatSwift/Plugin.swift
```diff
@@ -166,3 +167,92 @@ enum CommandError: Error {
...
+
+    while let fileURL = enumerator?.nextObject() as? URL {
+      let fileName = fileURL.lastPathComponent
+
+      guard fileName.hasPrefix("Package"), fileName.hasSuffix(".swift") else { continue }
+
+      // Parse the Swift version from the file name if it has a format like `Package@swift-5.8.swift`
+      if
+        fileName.hasPrefix("Package@swift-"),
+        let swiftVersion = parseSwiftVersion(from: fileName.dropFirst("Package@swift-".count))
```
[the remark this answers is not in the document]
calda: Here's some documentation about the `@swift-5.8` syntax: https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#version-specific-manifest-selection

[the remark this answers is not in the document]
calda: I suppose there's no reason to parse both the file name and `swift-tools-version`

Plugins/FormatSwift/Plugin.swift
```diff
@@ -166,3 +167,92 @@ enum CommandError: Error {
...
+      {
+        supportedSwiftVersions.append(swiftVersion)
+      }
+
+      // Parse the Swift tools version from the file body if it starts with a comment like `// swift-tools-version: 5.8`
```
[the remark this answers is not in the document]
calda: Good idea

Plugins/FormatSwift/Plugin.swift
```diff
@@ -166,3 +167,92 @@ enum CommandError: Error {
...
+    return supportedSwiftVersions
+  }
+
+  /// Parses a Swift version from a string like "5.8"
+  private func parseSwiftVersion(from string: Substring) -> SwiftVersion? {
+    var string = string
+
+    guard
+      string.count >= 3,
+      let major = string.popFirst().flatMap({ Int("\($0)") }),
```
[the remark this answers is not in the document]
calda: Perhaps this would be a good style guide rule 😄  (would be easy to autoformat)

Plugins/FormatSwift/Plugin.swift
bachand: This is great investigation. Thank you @calda .

bachand: Your findings have made me think that in general we should prefer to write code in the `AirbnbSwiftFormatTool` rather than the plugin. That being said, I see how it could complex to move this code into `AirbnbSwiftFormatTool` given context that only exists in the plugin. https://github.com/airbnb/swift/blob/8bd404ca686b72c1931d434f5286b6975833cc77/Plugins/FormatSwift/Plugin.swift#L210

Plugins/FormatSwift/Plugin.swift
```diff
@@ -92,9 +92,9 @@ extension AirbnbSwiftFormatPlugin: CommandPlugin {
       inputPaths = try self.inputPaths(for: context.package)
     }
 
-    // When running on a SPM package we infer the minimum Swift version from the
-    // `swift-tools-version` in `Package.swift` by default if the user doesn't
-    // specify one manually
+    // When running on a SPM package, by default we use the minimum swift version
+    // specified in any of the package manifest files. Alternatively the user can
+    // manually specify a swift version via the `--swift-version` argument.
```
bachand, line 97: Nice

Plugins/FormatSwift/Plugin.swift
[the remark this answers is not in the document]
bachand: > Perhaps this would be a good style guide rule 😄 (would be easy to autoformat)

This would be cool, but do you think we will have the type information we need to know if a `String` initializer exists?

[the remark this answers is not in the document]
calda: Good point

PR #236, opened 2023-09-05
Define default Swift compiler flags, automatically update Package.swift

calda:
Running CI on these draft changes, will add a PR description soon.


Plugins/FormatSwift/Plugin.swift
```diff
@@ -166,3 +169,92 @@ enum CommandError: Error {
   case lintFailure
   case unknownError(exitCode: Int32)
 }
+
+// MARK: - SwiftVersion
+
+struct SwiftVersion: Comparable {
```
calda, line 175: The changes in this PR depend on https://github.com/airbnb/swift/pull/235 for CI to pass

Sources/AirbnbSwiftFormatTool/SwiftSettings.swift
```diff
@@ -0,0 +1,114 @@
...
+      let originalContent = contents
+
+      let extensionRegex = Regex {
+        "extension [SwiftSetting] {"
+        OneOrMore { .any }
+        "static func airbnbDefault"
+        OneOrMore { .any }
+        OneOrMore { .newlineSequence }
+        "}"
+      }
```
calda, line 61: First time using the new Regex API!

Sources/AirbnbSwiftFormatTool/SwiftSettings.swift
```diff
@@ -0,0 +1,114 @@
...
+
+      contents.replace(extensionRegex, with: airbnbDefaultSwiftSettings)
```
calda, line 63: First time using the new Regex API! Really nice

Plugins/FormatSwift/Plugin.swift
calda: I'll rebase and remove these changes from this PR once https://github.com/airbnb/swift/pull/235 is merged.

Sources/AirbnbSwiftFormatTool/SwiftSettings.swift
```diff
@@ -0,0 +1,107 @@
...
+  static func updatePackageManifests(in packageDirectory: URL, lintOnly: Bool, verbose: Bool) throws {
+    for packageManifestURL in packageManifestURLs(in: packageDirectory) {
+      var contents = try String(contentsOf: packageManifestURL)
+      let originalContent = contents
+
+      let extensionRegex = Regex {
+        "extension [SwiftSetting] {"
+        OneOrMore { .any }
+        "static func airbnbDefault"
+        OneOrMore { .any }
```
calda, line 51: I wanted to use `/.*/` here, to demonstrate that we're able to use regex literals now in this module. This causes SwiftLint to fail with a parsing error though. In a follow up I can look in to updating to a SwiftLint version that supports this syntax.

jqsilver (approved):

Package.swift
```diff
@@ -43,8 +43,8 @@ let package = Package(
 
     .binaryTarget(
       name: "SwiftLintBinary",
-      url: "https://github.com/realm/SwiftLint/releases/download/0.48.0/SwiftLintBinary-macos.artifactbundle.zip",
-      checksum: "9c255e797260054296f9e4e4cd7e1339a15093d75f7c4227b9568d63edddba50"),
+      url: "https://github.com/realm/SwiftLint/releases/download/0.52.1/SwiftLintBinary-macos.artifactbundle.zip",
+      checksum: "bb4875e7a0a80b4799211f2eb35d4a81a9d4fc9175f06be4479a680d76ddf29c"),
```
calda, line 47: This portion of the diff will go away after we rebase on top of https://github.com/airbnb/swift/pull/239

Sources/AirbnbSwiftFormatTool/SwiftSettings.swift
calda: Fixed with https://github.com/airbnb/swift/pull/236/commits/7ddea2564f901b08e50b24bf5821c2cc0ec281b5 and https://github.com/airbnb/swift/pull/236/commits/e080364df3cd28d37a58812829de3fd8e9d63b90

calda:
Updated a few things:
 - Added unit tests which run in CI for the new functionality
 -  Added documentation to the readme

calda:
We've decided not to publish these flags as a part of our style guide.

One motivation of doing this is that it would let us define new style guide rules that depend on these set of flags being enabled. To do this we'd also have to list these flags in `airbnb.swiftformat`. That would make it so consumers of our package plugin _must_ have these build flags enabled, since our SwiftFormat config would think they did. Since enabling these flags is a breaking change for existing Swift 5 code, this would make it too difficult for consumers to use our style guide.

bachand:
Thanks @calda for putting up this PR and for generally keeping the style guide fresh!

In addition to what you shared about our reasoning for closing this PR, I want to add that we still are open to adding rules for new features of Swift. We can gate those rules in SwiftFormat on a particular Swift version. That way, consumers of our formatting tool/plugin will only be affected by the rules relevant to the version of Swift that they are using in their project, even if airbnb.swiftformat contains rules for newer versions of Swift.

If/when we add rules that only apply starting in a specific version of Swift, we can also consider mentioning that detail in the prose of the rule.

.github/workflows/main.yml
```diff
@@ -29,7 +29,16 @@ jobs:
...
         - '14.2' # Swift 5.7
         - '14.3' # Swift 5.8
+        - '15.0' # Swift 5.9
     steps:
       - uses: actions/checkout@v2
       - name: Test Package Plugin
-        run: swift package --allow-writing-to-package-directory format --lint
+        run: swift package --allow-writing-to-package-directory format --lint --log
+
+  unit-tests:
```
bachand, line 38: We can keep this code in mind if we want to add unit tests to this project in the future.

PR #240, opened 2023-09-25
Improve discussion and examples for rule to avoid unowned captures

calda:
#### Summary

This PR improves the discussion and examples for the "avoid `unowned` self" rule:
 - The examples now use `unowned self` and `weak self`, which feels less contrived than the previous examples
 - We now include discussion around explicitly capturing the instance members used in the closure (e.g. `[planet]` instead of `[weak self]` / `self?.planet`). 

----

`unowned` captures are unsafe because they will cause the application to crash if the referenced object has been deallocated.

```swift
// WRONG: Crashes if `self` has been deallocated when closures are called.
final class SpaceshipNavigationService {
  let spaceship: Spaceship
  let planet: Planet
  
  func colonizePlanet() {
    spaceship.travel(to: planet, onArrival: { [unowned self] in
      planet.colonize()
    })
  }
  
  func exploreSystem() {
    spaceship.travel(to: planet, nextDestination: { [unowned self] in
      planet.moons?.first
    })
  }
}
```

`weak` captures are safer because they require the author to explicitly handle the case where the referenced object no longer exists.

```swift
// RIGHT: Uses a `weak self` capture and explicitly handles the case where `self` has been deallocated
final class SpaceshipNavigationService {
  let spaceship: Spaceship
  let planet: Planet
  
  func colonizePlanet() {
    spaceship.travel(to: planet, onArrival: { [weak self] in
      guard let self else { return }
      planet.colonize()
    })
  }
  
  func exploreSystem() {
    spaceship.travel(to: planet, nextDestination: { [weak self] in
      guard let self else { return nil }
      return planet.moons?.first
    })
  }
}
```

Alternatively, consider directly capturing the variables that are used in the closure. This lets you avoid having to handle the case where `self` is nil, since you don't even need to reference `self`:

```swift
// RIGHT: Explicitly captures `planet` instead of capturing `self`
final class SpaceshipNavigationService {
  let spaceship: Spaceship
  let planet: Planet
  
  func colonizePlanet() {
    spaceship.travel(to: planet, onArrival: { [planet] in
      planet.colonize()
    })
  }
  
  func exploreSystem() {
    spaceship.travel(to: planet, nextDestination: { [planet] in
      planet.moons?.first
    })
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1622,50 +1622,75 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='unowned-captures'></a>(<a href='#unowned-captures'>link</a>) **Prefer using `weak` captures over `unowned` captures.**  [![SwiftLint: unowned_variable_capture](https://img.shields.io/badge/SwiftLint-unowned__variable__capture-007A87.svg)](https://realm.github.io/SwiftLint/unowned_variable_capture.html)
+* <a id='unowned-captures'></a>(<a href='#unowned-captures'>link</a>) **Avoid using `unowned` captures.** Instead prefer safer alternatives like `weak` captures, or capturing variables directly. [![SwiftLint: unowned_variable_capture](https://img.shields.io/badge/SwiftLint-unowned__variable__capture-007A87.svg)](https://realm.github.io/SwiftLint/unowned_variable_capture.html)
 
   <details>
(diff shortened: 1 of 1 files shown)


jqsilver (approved):
LGTM!

PR #242, opened 2023-09-25
Fix issue where package plugin wouldn't run SwiftLint lint-only rules unless passing in `--lint` flag

calda:
#### Summary

This PR fixes an issue where the package plugin wouldn't run any of the SwiftLint lint-only rules unless we passed in the `--lint` flag (e.g. `swift format format --lint`).

That is, running `swift package format` would only run the SwiftLint rules that included autocorrect, but wouldn't run any of the SwiftLint rules lacking autocorrect.

For SwiftLint to apply autocorrection, you have to pass in the `--fix` flag, but that causes SwiftFormat to not run any of the lint-only rules. As a fix we _always_ run SwiftLint in lint-only mode (without `--fix`) and then additionally run it in autocorrect mode (with `--fix`) if there were lint errors that need to be corrected.

This PR also adds unit tests for the `AirbnbSwiftFormatTool` type, testing its behavior in response to receiving various exit codes from the underlying commands.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,5 +45,5 @@ jobs:
         - '15.0' # Swift 5.9
     steps:
       - uses: actions/checkout@v2
-      - name: Test Package Plugin
+      - name: Run Unit Tests
         run: swift test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,7 +29,21 @@ jobs:
         xcode:
         - '14.2' # Swift 5.7
         - '14.3' # Swift 5.8
+        - '15.0' # Swift 5.9
     steps:
       - uses: actions/checkout@v2
       - name: Test Package Plugin
         run: swift package --allow-writing-to-package-directory format --lint
+
+  unit-tests:
+    name: Unit Tests
+    runs-on: macos-13
+    strategy:
+      fail-fast: false
+      matrix:
+        xcode:
+        - '15.0' # Swift 5.9
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test Package Plugin
+        run: swift test
--- a/README.md
+++ b/README.md
@@ -65,6 +65,10 @@ $ swift package format --targets AirbnbSwiftFormatTool
 $ swift package format --swift-version 5.3
 ```
 
+The package plugin returns a non-zero exit code if there is a lint failure that requires attention.
+ - In `--lint` mode, any lint failure from any tool will result in a non-zero exit code.
+ - In standard autocorrect mode without `--lint`, only failures from SwiftLint lint-only rules will result in a non-zero exit code.
+
 </details>
 
 ## Table of Contents
--- a/Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
+++ b/Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
@@ -40,40 +40,64 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
   @Option(help: "The project's minimum Swift version")
   var swiftVersion: String?
 
-  mutating func run() throws {
-    try swiftFormat.run()
-    swiftFormat.waitUntilExit()
+  func run() throws {
+    let swiftFormat = makeSwiftFormatCommand()
+    let swiftLint = makeSwiftLintCommand(autocorrect: false)
+    let swiftLintAutocorrect = makeSwiftLintCommand(autocorrect: true)
 
-    try swiftLint.run()
-    swiftLint.waitUntilExit()
+    let swiftFormatExitCode = try swiftFormat.run()
 
-    if log {
-      log(swiftFormat.shellCommand)
-      log(swiftLint.shellCommand)
-      log("SwiftFormat ended with exit code \(swiftFormat.terminationStatus)")
-      log("SwiftLint ended with exit code \(swiftLint.terminationStatus)")
+    // We always have to run SwiftLint in lint-only mode at least once,
+    // because when in autocorrect mode SwiftLint won't emit any lint warnings.
+    let swiftLintExitCode = try swiftLint.run()
+
+    let swiftLintAutocorrectExitCode: Int32?
+    if
+      // When only linting, we shouldn't run SwiftLint with autocorrect enabled
+      !lintOnly,
+      // When formatting, only re-run SwiftLint a second time if the first
+      // invocation reported a lint failure that needs to be corrected
+      swiftLintExitCode == SwiftLintExitCode.lintFailure
+    {
+      swiftLintAutocorrectExitCode = try swiftLintAutocorrect.run()
+    } else {
+      swiftLintAutocorrectExitCode = nil
     }
 
     if
-      swiftFormat.terminationStatus == SwiftFormatExitCode.lintFailure ||
-      swiftLint.terminationStatus == SwiftLintExitCode.lintFailure
+      swiftFormatExitCode == SwiftFormatExitCode.lintFailure ||
+      swiftLintExitCode == SwiftLintExitCode.lintFailure ||
+      swiftLintAutocorrectExitCode == SwiftLintExitCode.lintFailure
     {
       throw ExitCode.failure
     }
 
     // Any other non-success exit code is an unknown failure
-    if swiftFormat.terminationStatus != EXIT_SUCCESS {
-      throw ExitCode(swiftFormat.terminationStatus)
+    if swiftFormatExitCode != EXIT_SUCCESS {
+      throw ExitCode(swiftFormatExitCode)
     }
 
-    if swiftLint.terminationStatus != EXIT_SUCCESS {
-      throw ExitCode(swiftLint.terminationStatus)
+    if swiftLintExitCode != EXIT_SUCCESS {
+      throw ExitCode(swiftLintExitCode)
+    }
+
+    if
+      let swiftLintAutocorrectExitCode = swiftLintAutocorrectExitCode,
+      swiftLintAutocorrectExitCode != EXIT_SUCCESS
+    {
+      throw ExitCode(swiftLintAutocorrectExitCode)
     }
   }
 
   // MARK: Private
 
-  private lazy var swiftFormat: Process = {
+  /// Whether the command should autocorrect invalid code, or only emit lint errors
+  private var lintOnly: Bool {
+    lint
+  }
+
+  /// Builds a command that runs the SwiftFormat tool
+  private func makeSwiftFormatCommand() -> Command {
     var arguments = directories + [
       "--config", swiftFormatConfig,
     ]
@@ -90,13 +114,16 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
       arguments += ["--swiftversion", swiftVersion]
     }
 
-    let swiftFormat = Process()
-    swiftFormat.launchPath = swiftFormatPath
-    swiftFormat.arguments = arguments
-    return swiftFormat
-  }()
+    return Command(
+      log: log,
+      launchPath: swiftFormatPath,
+      arguments: arguments)
+  }
 
-  private lazy var swiftLint: Process = {
+  /// Builds a command that runs the SwiftLint tool
+  ///  - If `autocorrect` is true, passes the `--fix` flag to SwiftLint.
+  ///    When autocorrecting, SwiftLint doesn't emit any lint warnings.
+  private func makeSwiftLintCommand(autocorrect: Bool) -> Command {
     var arguments = directories + [
       "--config", swiftLintConfig,
       // Required for SwiftLint to emit a non-zero exit code on lint failure
@@ -107,15 +134,15 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
       arguments += ["--cache-path", swiftLintCachePath]
     }
 
-    if !lint {
+    if autocorrect {
       arguments += ["--fix"]
     }
 
-    let swiftLint = Process()
-    swiftLint.launchPath = swiftLintPath
-    swiftLint.arguments = arguments
-    return swiftLint
-  }()
+    return Command(
+      log: log,
+      launchPath: swiftLintPath,
+      arguments: arguments)
+  }
 
   private func log(_ string: String) {
     // swiftlint:disable:next no_direct_standard_out_logs
--- a/Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
+++ b/Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
@@ -47,22 +47,21 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
 
     let swiftFormatExitCode = try swiftFormat.run()
 
-    // We always have to run SwiftLint in lint-only mode at least once,
-    // because when in autocorrect mode SwiftLint won't emit any lint warnings.
-    let swiftLintExitCode = try swiftLint.run()
-
+    // Run SwiftLint in autocorrect mode first, so that if autocorrect fixes all of the SwiftLint violations
+    // then the following lint-only invocation will not report any violations.
     let swiftLintAutocorrectExitCode: Int32?
     if
       // When only linting, we shouldn't run SwiftLint with autocorrect enabled
-      !lintOnly,
-      // When formatting, only re-run SwiftLint a second time if the first
-      // invocation reported a lint failure that needs to be corrected
-      swiftLintExitCode == SwiftLintExitCode.lintFailure
+      !lintOnly
     {
       swiftLintAutocorrectExitCode = try swiftLintAutocorrect.run()
     } else {
       swiftLintAutocorrectExitCode = nil
     }
+
+    // We always have to run SwiftLint in lint-only mode at least once,
+    // because when in autocorrect mode SwiftLint won't emit any lint warnings.
+    let swiftLintExitCode = try swiftLint.run()
 
     if
       swiftFormatExitCode == SwiftFormatExitCode.lintFailure ||
--- a/Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
+++ b/Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
@@ -89,6 +89,16 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
     }
   }
 
+  /// Runs the `AirbnbSwiftFormatTool` command with the given closure that executes each individual command.
+  func run(executeCommand: @escaping (Command) -> Int32) throws {
+    let existingImplementation = Command.runCommand
+
+    Command.runCommand = executeCommand
+    defer { Command.runCommand = existingImplementation }
+
+    try run()
+  }
+
   // MARK: Private
 
   /// Whether the command should autocorrect invalid code, or only emit lint errors
--- a/Sources/AirbnbSwiftFormatTool/Command.swift
+++ b/Sources/AirbnbSwiftFormatTool/Command.swift
@@ -9,18 +9,22 @@ struct Command {
   // MARK: Internal
 
   /// Mock implementation of `Command.run` which can be provided during unit test
-  static var _mockRunCommand: ((Command) -> Int32)?
+  static var runCommand: (Command) throws -> Int32 = { try $0.executeShellCommand() }
 
   let log: Bool
   let launchPath: String
   let arguments: [String]
 
-  /// Synchronously runs this command and returns its exit code
+  /// Runs this command using the implementation of `Command.runCommand`
+  ///  - By default, synchronously runs this command and returns its exit code
   func run() throws -> Int32 {
-    if let _mockRunCommand = Command._mockRunCommand {
-      return _mockRunCommand(self)
-    }
+    try Command.runCommand(self)
+  }
 
+  // MARK: Private
+
+  /// Synchronously runs this command and returns its exit code
+  private func executeShellCommand() throws -> Int32 {
     let process = Process()
     process.launchPath = launchPath
     process.arguments = arguments
@@ -39,8 +43,6 @@ struct Command {
 
     return process.terminationStatus
   }
-
-  // MARK: Private
 
   private func log(_ string: String) {
     // swiftlint:disable:next no_direct_standard_out_logs
--- a/Sources/AirbnbSwiftFormatTool/Command.swift
+++ b/Sources/AirbnbSwiftFormatTool/Command.swift
@@ -8,7 +8,7 @@ struct Command {
 
   // MARK: Internal
 
-  /// Mock implementation of `Command.run` which can be provided during unit test
+  /// This property can be overridden to provide a mock implementation in unit tests.
   static var runCommand: (Command) throws -> Int32 = { try $0.executeShellCommand() }
 
   let log: Bool
--- a/Sources/AirbnbSwiftFormatTool/Command.swift
+++ b/Sources/AirbnbSwiftFormatTool/Command.swift
@@ -1,0 +1,50 @@
+// Created by Cal Stephens on 9/25/23.
+// Copyright © 2023 Airbnb Inc. All rights reserved.
+
+import Foundation
+
+/// A single command line invocation
+struct Command {
+
+  // MARK: Internal
+
+  /// Mock implementation of `Command.run` which can be provided during unit test
+  static var _mockRunCommand: ((Command) -> Int32)?
+
+  let log: Bool
+  let launchPath: String
+  let arguments: [String]
+
+  /// Synchronously runs this command and returns its exit code
+  func run() throws -> Int32 {
+    if let _mockRunCommand = Command._mockRunCommand {
+      return _mockRunCommand(self)
+    }
+
+    let process = Process()
+    process.launchPath = launchPath
+    process.arguments = arguments
+
+    if log {
+      log(process.shellCommand)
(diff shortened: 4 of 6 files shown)


calda:
ptal @bachand

Sources/AirbnbSwiftFormatTool/Command.swift
```diff
@@ -0,0 +1,50 @@
...
+
+  let log: Bool
+  let launchPath: String
+  let arguments: [String]
+
+  /// Synchronously runs this command and returns its exit code
+  func run() throws -> Int32 {
+    if let _mockRunCommand = Command._mockRunCommand {
+      return _mockRunCommand(self)
+    }
```
bachand, line 22: In general I think it's preferable if we can avoid having test-only code in the production implementation. I am curious if there is a way for us to avoid the need for `_mockRunCommand`. To that end, I want to propose an idea...

What if `Command` was a value type that encoded all of the information about the command we wanted to run, but didn't actually run the command. Then there could be separate logic in `AirbnbSwiftFormatTool` to run the command.

We could make this more fancy, but at a high level I am thinking that we can initialize `AirbnbSwiftFormatTool` with a way to run the command.

```swift
struct AirbnbSwiftFormatTool {
  init(runCommand: @escaping (Command) -> Int32 = Self.defaultRunCommand(command:)) {
    self.runCommand = runCommand
  }
    
  static func defaultRunCommand(command: Command) -> Int32 {
    // contains an adaptation of the `run()` method on `Command` as it exists now
  }
    
  private let runCommand: (Command) -> Int32
}
```

Then we can pass in an alternate closure for `runCommand` when initializing  `AirbnbSwiftFormatTool` under test.

Granted, we are still adapting production to be a bit more complex, but IMHO the injection point we are creating is cleaner as it's less tied to unit testing. For example, I think this pattern would naturally lend itself to supporting a `--dry-run` option, potentially by injecting a command runner object instead of a closure.

```swift
protocol CommandRunner {
  func runCommand(_ command: Command) -> Int32
}

class DefaultCommandRunner: CommandRunner {

  func runCommand(command: Command) -> Int32 {
    // contains an adaptation of the `run()` method on `Command` as it exists now
  }
}

class DryRunDefaultCommandRunner: CommandRunner {

  func runCommand(command: Command) -> Int32 {
    // prints out the command instead of running it
  }
}

class MockCommandRunner: CommandRunner { … }

struct AirbnbSwiftFormatTool {
  init(
    commandRunner: CommandRunner = DefaultCommandRunner(),
    dryRunCommandRunner: CommandRunner = DryRunDefaultCommandRunner())
  {
    self.commandRunner = commandRunner
    self.dryRunCommandRunner = dryRunCommandRunner
  }
  
  func run() throws {
    // chooses `commandRunner` or `dryRunCommandRunner` based on the `--dry-run` option
  }

  private let commandRunner: CommandRunner
  private let dryRunCommandRunner: CommandRunner
}
```

Then we could initialize the `AirbnbSwiftFormatTool` as follows in a unit test environment.

```swift
let mockCommandRunner = MockCommandRunner()
// configure mockCommandRunner
let mockDryRunCommandRunner = MockCommandRunner()
// configure mockDryRunCommandRunner

let tool = AirbnbSwiftFormatTool(
  commandRunner: mockCommandRunner,
  dryRunCommandRunner: mockDryRunCommandRunner)

// run tool
// assert on state of mockCommandRunner and mockDryRunCommandRunner
```

bachand (commented):
Thanks @calda (and sorry for the delay!!). I am curious to see what you think about my comment on how we could approach enabling testability in a different way.

.github/workflows/main.yml
```diff
@@ -29,7 +29,21 @@ jobs:
...
+    name: Unit Tests
+    runs-on: macos-13
+    strategy:
+      fail-fast: false
+      matrix:
+        xcode:
+        - '15.0' # Swift 5.9
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test Package Plugin
```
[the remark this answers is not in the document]
calda: Thanks

Sources/AirbnbSwiftFormatTool/Command.swift
calda: Sure, seems reasonable

calda: I took a look and this doesn't appear possible for a few reasons:
 1. We can't call `AirbnbSwiftFormatTool.init` directly. `ArgumentParser` `ParsableCommand` implementations must be initialized via the `static func parse` method in order to correctly set up the instance properties, which are all property wrappers provided by the `ArgumentParser` library.
      - This is not a huge obstacle because we could work around it by just making the `runCommand` property a `var` instead of a `let` and just customize it after calling `init`.
 2. We can't add dependencies like this as instance members of `AirbnbSwiftFormatTool` since `ParsableCommand` must be `Decodable`. I can't think of a way to add a closure or builder-like object as an instance member of `AirbnbSwiftFormatTool` without doing something unusual (like making the builder conform to `Decodable` and having that just populate the default implementation) or manually implementing the `decode` method (adds overhead and doesn't seem like idiomatic usage of the `ArgumentParser` library).

It seems this points us to needing to use a global `static var` closure. 

Do you think it would be preferable to use something like:

```swift
static var runCommand: (Command) -> Int32 = { try $0.run() }
```

which is always called even in production, and can be customized for unit tests, over the current design?

bachand: Thanks for investigating @calda . What you wrote above makes sense.

I think the static function could work well. Another path could be moving more of the code in `AirbnbSwiftFormatTool` into another object that doesn't have so many limitations (e.g. a `Formatter` object). Then `AirbnbSwiftFormatTool`'s `run()` method could instantiate a `Formatter` and call into that object to do the bulk of the work. In this alternate path, we would write the unit tests on this new object.

I can get onboard with either path.

calda: I'll go with the static function approach so we can continue using the existing code structure, which seems to be the idiomatic way to use the `ArgumentParser` library.

calda: Here you go: https://github.com/airbnb/swift/pull/242/commits/c13dd242f6ad319117002e0c8a859d453bc7654b

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -40,40 +40,74 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
+    try run()
+  }
+
   // MARK: Private
 
-  private lazy var swiftFormat: Process = {
+  /// Whether the command should autocorrect invalid code, or only emit lint errors
+  private var lintOnly: Bool {
+    lint
+  }
```
bachand, line 107: What do you think about updating the `lint` property to be called `lintOnly`? This would be a breaking API change but I think it would increase the clarity of the CLI interface too. `--lint-only` seems more clear than `--lint` if the default behavior is autocorrection.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -40,40 +40,74 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
+    } else {
+      swiftLintAutocorrectExitCode = nil
     }
 
     if
-      swiftFormat.terminationStatus == SwiftFormatExitCode.lintFailure ||
-      swiftLint.terminationStatus == SwiftLintExitCode.lintFailure
+      swiftFormatExitCode == SwiftFormatExitCode.lintFailure ||
+      swiftLintExitCode == SwiftLintExitCode.lintFailure ||
+      swiftLintAutocorrectExitCode == SwiftLintExitCode.lintFailure
```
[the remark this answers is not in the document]
bachand: I ran tests locally and this is what I found as some relevant data points for the exit code when there is a lint failure.

| Tool | Lint-only command | Exit code |
| ---- | ----- | ----- |
| SwiftFormat | `swiftformat dir/with/lint/failure --config airbnb.swiftformat --lint` | 1 |
| SwiftLint | `swiftlint dir/with/lint/failure --config swiftlint.yml --strict` | 2 |

| Tool | Autocorrect command | Exit code |
| ---- | ----- | ----- |
| SwiftFormat | `swiftformat dir/with/lint/failure --config airbnb.swiftformat` | 0 |
| SwiftLint | `swiftlint dir/with/lint/failure --config swiftlint.yml --strict --fix` | 0 |

I ran these tests using this lint failure, which fails both linter and is autocorrect by both linters.
Screenshot 2023-10-11 at 3 08 07 PM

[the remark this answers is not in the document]
bachand: This ^ behavior seems reasonable to me and seems like how our tool should function as well. What do you think about documenting the exit code behavior of this tool in the README as part of this PR and then we can ensure that the behavior matches the documented behavior. Right now AFAICT we will fail with a nonzero exit code if there is a SwiftLint error that we autocorrect, which seems undesirable.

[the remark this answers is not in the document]
calda: That seems reasonable. What do you think about making that change in a follow-up since (I think?) it's a separate behavior change from the one in this PR.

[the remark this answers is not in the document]
calda: It seems you're probably right that we throw a non-zero exit code when autocorrecting, which is probably unexpected given that the underlying tools always return a zero exit code when autocorrecting.

README.md
```diff
@@ -65,6 +65,10 @@ $ swift package format --targets AirbnbSwiftFormatTool
 $ swift package format --swift-version 5.3
 ```
 
+The package plugin returns a non-zero exit code if there is a lint failure that requires attention.
+ - In `--lint` mode, any lint failure from any tool will result in a non-zero exit code.
+ - In standard autocorrect mode without `--lint`, only failures from SwiftLint lint-only rules will result in a non-zero exit code.
```
bachand, line 70: @calda do you know if there is the possibility of a situation where there is an autocorrectable rule that can't be autocorrected in all cases (and thus would continue to result in a non-zero exit code after autocorrection)?

README.md
```diff
@@ -65,6 +65,10 @@ $ swift package format --targets AirbnbSwiftFormatTool
 $ swift package format --swift-version 5.3
 ```
 
+The package plugin returns a non-zero exit code if there is a lint failure that requires attention.
```
bachand, line 68: Thank you for adding.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -40,40 +40,73 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
-      log("SwiftLint ended with exit code \(swiftLint.terminationStatus)")
+  func run() throws {
+    let swiftFormat = makeSwiftFormatCommand()
+    let swiftLint = makeSwiftLintCommand(autocorrect: false)
+    let swiftLintAutocorrect = makeSwiftLintCommand(autocorrect: true)
+
+    let swiftFormatExitCode = try swiftFormat.run()
+
+    // Run SwiftLint in autocorrect mode first, so that if autocorrect fixes all of the SwiftLint violations
+    // then the following lint-only invocation will not report any violations.
```
bachand, line 51: 👍

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: I think `--lint` is alright. SwiftFormat uses `swiftformat --lint` to mean "only lint, don't apply any changes" as an example.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -40,40 +40,73 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
+      !lintOnly
+    {
+      swiftLintAutocorrectExitCode = try swiftLintAutocorrect.run()
+    } else {
+      swiftLintAutocorrectExitCode = nil
     }
 
+    // We always have to run SwiftLint in lint-only mode at least once,
+    // because when in autocorrect mode SwiftLint won't emit any lint warnings.
+    let swiftLintExitCode = try swiftLint.run()
```
bachand, line 64: 👍

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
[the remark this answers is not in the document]
calda: https://github.com/airbnb/swift/pull/242/commits/46618a1d73d5e940fde2cbba6e008eae12464519

Sources/AirbnbSwiftFormatTool/Command.swift
```diff
@@ -0,0 +1,52 @@
...
+// Copyright © 2023 Airbnb Inc. All rights reserved.
+
+import Foundation
+
+/// A single command line invocation
+struct Command {
+
+  // MARK: Internal
+
+  /// Mock implementation of `Command.run` which can be provided during unit test
```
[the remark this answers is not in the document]
calda: Thanks

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: If we were starting from scratch I'd be cool with the `--lint-only` suggestion but given we already have consumers of this tool I'd prefer to not make breaking changes

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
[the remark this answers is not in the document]
bachand: > Although, what exit code would we use in a case where:
> 
> 1. We are not in lint-only mode (implying we should always return a 0 exit code?)
> 2. A SwiftLint lint-only rule triggers a failure (here that SwiftLint invocation would return a non-zero exit code)

We aligned that we _do_ want our tool to return a non-zero exit code when there is a lint-only failure even if the tool is run in autocorrect mode and the autocorrectable violations are fixed.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
bachand: sg

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -40,40 +40,73 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
+      throw ExitCode(swiftFormatExitCode)
     }
 
-    if swiftLint.terminationStatus != EXIT_SUCCESS {
-      throw ExitCode(swiftLint.terminationStatus)
+    if swiftLintExitCode != EXIT_SUCCESS {
+      throw ExitCode(swiftLintExitCode)
+    }
+
+    if
```
bachand, line 83: 👍

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -40,40 +40,73 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
+      let swiftLintAutocorrectExitCode = swiftLintAutocorrectExitCode,
+      swiftLintAutocorrectExitCode != EXIT_SUCCESS
+    {
+      throw ExitCode(swiftLintAutocorrectExitCode)
     }
   }
 
+  /// Runs the `AirbnbSwiftFormatTool` command with the given closure that executes each individual command.
```
bachand, line 91: @calda what do you think about inlining this into `AirbnbSwiftFormatToolTest`'s `runFormatTool(…)` method? Then we avoid shipping this test-only code in our production binary.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -90,13 +123,16 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
+    return Command(
+      log: log,
+      launchPath: swiftFormatPath,
+      arguments: arguments)
+  }
 
-  private lazy var swiftLint: Process = {
+  /// Builds a command that runs the SwiftLint tool
+  ///  - If `autocorrect` is true, passes the `--fix` flag to SwiftLint.
+  ///    When autocorrecting, SwiftLint doesn't emit any lint warnings.
```
bachand, line 134: 👍

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -90,13 +123,16 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
       arguments += ["--swiftversion", swiftVersion]
     }
 
-    let swiftFormat = Process()
-    swiftFormat.launchPath = swiftFormatPath
-    swiftFormat.arguments = arguments
-    return swiftFormat
-  }()
+    return Command(
```
bachand, line 126:
```suggestion
    return .init(
```
Nit: I think this is preferable since it makes refactoring easier. Could be a style guide rule one day :)

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -107,15 +143,15 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
...
+    if autocorrect {
       arguments += ["--fix"]
     }
 
-    let swiftLint = Process()
-    swiftLint.launchPath = swiftLintPath
-    swiftLint.arguments = arguments
-    return swiftLint
-  }()
+    return Command(
```
bachand, line 150: nit
```suggestion
    return .init(
```

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
```diff
@@ -0,0 +1,255 @@
...
+        swiftLintAutocorrect: {
+          ranSwiftLintAutocorrect = true
+          return EXIT_SUCCESS
+        }))
+
+    XCTAssertNil(error)
+    XCTAssertTrue(ranSwiftFormat)
+    XCTAssertTrue(ranSwiftLint)
+
+    // Should't run SwiftLint autocorrect in lint-only mode
```
bachand, line 66: 👍

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
```diff
@@ -0,0 +1,255 @@
...
+        swiftFormat: {
+          ranSwiftFormat = true
+
+          // When autocorrecting SwiftFormat returns EXIT_SUCCESS
+          // even if there were violations that were fixed
+          return EXIT_SUCCESS
+        },
+        swiftLint: {
+          ranSwiftLint = true
+          return SwiftLintExitCode.lintFailure
```
bachand, line 86: 👍

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
```diff
@@ -0,0 +1,255 @@
...
+          // Assume that the codebase has violations that would be corrected by SwiftLint autocorrect.
+          if ranSwiftLintAutocorrect {
+            // If SwiftLint autocorrect has already run, then there are no more violations.
+            // This is the expected behavior.
+            return EXIT_SUCCESS
+          } else {
+            // If SwiftLint autocorrect hasn't run yet, then there are still violations.
+            // This should not happen, because we run autocorrect first.
+            return SwiftLintExitCode.lintFailure
+          }
```
bachand, line 125: Nice

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
```diff
@@ -0,0 +1,255 @@
...
+          return EXIT_SUCCESS
+        }))
+
+    XCTAssertEqual(error as? ExitCode, ExitCode.failure)
+    XCTAssertTrue(ranSwiftFormat)
+    XCTAssertTrue(ranSwiftLint)
+    XCTAssertFalse(ranSwiftLintAutocorrect)
+  }
+
+  func testLintWithOnlySwiftLintViolation() {
```
bachand, line 170: For completeness, should we have a test case for lint with only SwiftFormat violations?

bachand (approved):
SUPER great work @calda . Love it!

README.md
calda: Not sure. My guess is no but I don't know.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: Sure, thats good too

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: Not sure I agree -- I personally prefer using the non `.init(` form (with the type name) in the majority of cases because I find it makes the code easier to read. (I've seen lots of complex invocations with multiple `.init`s before where the code would be a lot easier to understand if the type names were written out explicitly instead of using the bare `.init`).

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
calda: I considered it, but the code for that test case would be identical to the existing `testLintWithNoViolations` test case (since in both cases SwiftFormat just returns `EXIT_SUCCESS`) so I'm not sure it adds much value.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: I can see "prefer .init where possible" being in alignment with our style guide principles (where we generally omit things that are unnecessary) but I find it a bit hard to swallow subjectively, and I think there are many cases where `.init` would read as less idiomatic.

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
calda: Oops, I made this change locally and then accidentally merged the PR before pushing. Do you think I should open a follow-up PR with this change?

bachand: Yeah if you can I think that would be great @calda since then we get the unit testing into a really solid state 👍

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
bachand: Thanks for sharing your POV on this. _If_ we adopted a "prefer .init where possible" rule it means that we need to rely more on IDE features for context, which can admittedly be finicky.

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
bachand: @calda I want to make sure we're on the same page. In the case where we run SwiftFormat in lint-only mode it returns a non-zero exit code (see https://github.com/airbnb/swift/pull/242#discussion_r1355831828). I am pretty sure that the new test case would be unique.

```swift
  func testLintWithOnlySwiftFormatViolation() {
    var ranSwiftFormat = false
    var ranSwiftLint = false
    var ranSwiftLintAutocorrect = false

    let error = runFormatTool(
      arguments: ["--lint"],
      with: MockCommands(
        swiftFormat: {
          ranSwiftFormat = true
          return SwiftFormatExitCode.lintFailure
        },
        swiftLint: {
          ranSwiftLint = true
          return EXIT_SUCCESS
        },
        swiftLintAutocorrect: {
          ranSwiftLintAutocorrect = true
          return EXIT_SUCCESS
        }))

    XCTAssertEqual(error as? ExitCode, ExitCode.failure)
    XCTAssertTrue(ranSwiftFormat)
    XCTAssertTrue(ranSwiftLint)
    XCTAssertFalse(ranSwiftLintAutocorrect)
  }
```

Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
bachand: (but also no worries if you don't have the bandwidth as we can do this in a future PR)

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
calda: My mistake, I misinterpreted your comment. It makes sense to add a test for this.

PR #244, opened 2023-10-18
Improve unit tests

calda:
This PR improves the unit tests added in #242, based on some feedback from @bachand that wasn't fully incorporated in the previous PR.


Sources/AirbnbSwiftFormatTool/AirbnbSwiftFormatTool.swift
```diff
@@ -88,16 +88,6 @@ struct AirbnbSwiftFormatTool: ParsableCommand {
     }
   }
 
-  /// Runs the `AirbnbSwiftFormatTool` command with the given closure that executes each individual command.
-  func run(executeCommand: @escaping (Command) -> Int32) throws {
```
calda, line 92: https://github.com/airbnb/swift/pull/242#discussion_r1355882206

PR #245, opened 2023-10-18
Soften wording for rule to avoid using default in switch statements

calda:
#### Summary

We have a rule that suggests to avoid using the `default` case when switching over an enum.

This rule is worded pretty strongly ("never use the `default` case"), but lacks autocorrect or linting, so has no enforcement. I also think there are valid reasons to use the default case.

I think we should just soften the wording of this rule (not use strict wording like "never"), since it is still good advice overall.

Some alternatives we could consider:
  1. Add a lint rule to enforce this rule
     - This is too heavy-handed, because there are valid use cases for using `default`. The prose rule includes a set of exceptions where it can make sense to use `default`.
     - This is infeasible, because a linter wouldn't know if we are switching over an enum or not. We may be switching over something that can't be switched over exhaustively, which would require a `default` case. (edit: Michael has suggested some heuristics we could use to reduce the number of false positives here without needing full type information).
  3. Remove this rule completely
     - We don't have many rules that are mere suggestions, without any linter enforcement or autocorrection. We haven't added any _new_ rules like this in many years.
     - I think we should continue to permit this sorts of rule, which makes recommendations that can result in higher-quality code but aren't strictly enforced. This is probably the biggest "expansion opportunity" for the content in our style guide.

#### Reasoning

### DIFF
--- a/README.md
+++ b/README.md
@@ -2289,7 +2289,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='switch-never-default'></a>(<a href='#switch-never-default'>link</a>) When switching over an enum, generally prefer enumerating all cases rather than using the `default` case.
+* <a id='switch-avoid-default'></a>(<a href='#switch-avoid-default'>link</a>) When switching over an enum, generally prefer enumerating all cases rather than using the `default` case.
 
   <details>
 
@@ -2322,7 +2322,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
     case success(Success)
     case failure(Error)
 
-    // We don't expect additional cases that may be added to the enumeration affecting this method.
+    // We expect that this property will remain valid if additional cases are added to the enumeration.
     public var isRunning: Bool {
       switch self {
       case .running:
@@ -2334,7 +2334,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
 
   extension TaskState: Equatable {
-    // Explicitly listing each state would be too burdensome. Ideally this function could be implemented with a custom macro.
+    // Explicitly listing each state would be too burdensome. Ideally this function could be implemented with a well-tested macro.
     public static func == (lhs: TaskState, rhs: TaskState) -> Bool {
       switch (lhs, rhs) {
       case (.pending, .pending):
--- a/README.md
+++ b/README.md
@@ -2313,9 +2313,45 @@ _You can enable the following settings in Xcode by running [this script](resourc
     // Stop your vehicle
   }
   
-  This is not always necessary, though:
-   - If the enum is very large, it may be unwieldy or unreasonable to enumerate each case in every switch statement.
-   - If you don't expect to need to exhaustively handle all enum cases added in the future at every call site, using `default` can reduce the overhead of introducing new enum cases.
+  // COUNTEREXAMPLES
+
+  enum TaskState {
+    case pending
+    case running
+    case canceling
+    case success(Success)
+    case failure(Error)
+
+    // We don't expect additional cases that may be added to the enumeration affecting this method.
+    public var isRunning: Bool {
+      switch self {
+      case .running:
+        true
+      default:
+        false
+      }
+    }  
+  }
+
+  extension TaskState: Equatable {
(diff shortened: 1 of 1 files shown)


bachand:
My opinion is that we should keep this rule worded as-is.

In a large codebase, it can be _very_ hard (I would argue infeasible) to validate all `switch` statements over an enumeration when adding a new case. This style guide rule makes that intractable validation problem trivial to validate, since the compiler does the work for us. IMHO the huge time savings and increase in confidence that this rule brings justifies the lack of enforcement and autocorrection. Our style guide indicates that exceptions should be heavily justified.

Screenshot 2023-10-18 at 12 07 13 PM

I think this is one of the cases where the benefits of an unenforceable rule outweigh the exceptionality of it.

README.md
```diff
@@ -2289,29 +2289,33 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
 
-* <a id='switch-never-default'></a>(<a href='#switch-never-default'>link</a>) **Never use the `default` case when `switch`ing over an enum.**
+* <a id='switch-never-default'></a>(<a href='#switch-never-default'>link</a>) When switching over an enum, generally prefer enumerating all cases rather than using the `default` case.
```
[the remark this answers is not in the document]
calda: Pragmatically, there are 6400+ `default:` cases in our codebase right now. I don't think we should remove all of these, or require that all of them manually suppress the linter. When auditing a reasonable number of these, I find a fair number that I think look totally reasonable as-is.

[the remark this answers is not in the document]
bachand: I think that _eventually_ there will be a Swift linter that is able to be run on the entire codebase and thus is able to have more complete knowledge. Such a linter would be more able to detect that a value is an enumeration, and thus avoid over-enforcing. If you believe that too, then I feel like we should be skating to where the world is moving rather than relaxing rules. And if we wanted, we could play a part in making a more advanced linter like that a reality :)

calda:
Worked with @bachand to workshop and simplify the changes

README.md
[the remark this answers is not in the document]
bachand: When chatting about this off of github.com, I also mentioned that I think it would be appropriate if you could configure the compiler to emit a warning when a `default:` case is used with a `switch` statement over an enumeration.

README.md
```diff
@@ -2289,29 +2289,69 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
```
bachand, line 2292: @calda I'm not sure what the right thing to do here is but do you think we should change the slug? 

```suggestion
* <a id='switch-avoid-default'></a>(<a href='#switch-avoid-default'>link</a>) When switching over an enum, generally prefer enumerating all cases rather than using the `default` case.
```

This change would break any links in the wild to the old wording. This may be what we want in this case, since we are intentionally softening this rule.

README.md
```diff
@@ -2289,29 +2289,69 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  // COUNTEREXAMPLES
+
+  enum TaskState {
+    case pending
+    case running
+    case canceling
+    case success(Success)
+    case failure(Error)
+
+    // We don't expect additional cases that may be added to the enumeration affecting this method.
```
bachand, line 2325:
```suggestion
    // We expect that this property will remain valid if additional cases are added to the enumeration.
```

README.md
```diff
@@ -2289,29 +2289,69 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+      case .running:
+        true
+      default:
+        false
+      }
+    }  
+  }
+
+  extension TaskState: Equatable {
+    // Explicitly listing each state would be too burdensome. Ideally this function could be implemented with a custom macro.
```
bachand, line 2337:
```suggestion
    // Explicitly listing each state would be too burdensome. Ideally this function could be implemented with a well-tested macro.
```

README.md
bachand: In some ways I think it would be ideal to have "prefer" in the new slug but I didn't think of a good way to do that.

bachand (approved):
🎉

README.md
calda: I think `switch-avoid-default` is good. I also think changing the slug is a good idea, thanks for the suggestion.

PR #247, opened 2023-10-26
Update Package.swift

elfredpagan:
#### Summary

Moving to the latest version

#### Reasoning

We use this version internally at airbnb.


calda (approved):
Thanks!

PR #248, opened 2023-10-27
Update to Swift 5.9

calda:
This PR updates the default SwiftFormat `--swiftversion` to Swift 5.9, which is the Swift version we now use internally at Airbnb.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -42,8 +42,8 @@ let package = Package(
 
     .binaryTarget(
       name: "SwiftFormat",
-      url: "https://github.com/calda/SwiftFormat/releases/download/0.53-beta-1/SwiftFormat.artifactbundle.zip",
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -42,8 +42,8 @@ let package = Package(
 
     .binaryTarget(
       name: "SwiftFormat",
-      url: "https://github.com/calda/SwiftFormat/releases/download/0.53-beta-1/SwiftFormat.artifactbundle.zip",
-      checksum: "a97a1231fd279d826c03611ec89772ee1131805f17f32bb16575bcd373d398a6"),
+      url: "https://github.com/calda/SwiftFormat/releases/download/0.53-beta-3/SwiftFormat.artifactbundle.zip",
```
bachand, line 45: I see that 0.53-beta-3 is still the latest version.

bachand (approved):
Thank you @calda !

PR #249, opened 2023-11-01
Update omit-return rule discussion to include examples using if/swift expressions

calda:
Now that we use [Swift 5.9](https://github.com/airbnb/swift/pull/248), the existing omit-return rule now also applies to functions / properties that have single-expression if/switch statements. Let's add some examples using this syntax to make it explicit that this existing rule applies to those cases. This is also potentially good educational content for anyone who doesn't know that this is permitted now.


bachand (approved):
Nice. This is great @calda !

PR #250, opened 2023-11-01
Add rule to prefer using if/switch expressions when initializing a new property with the result of a conditional statement

calda:
#### Summary

This PR proposes a new rule to prefer using if/switch expressions when initializing a new property with the result of a conditional statement. 

Support for this syntax was added in Swift 5.9, and autocorrect is implemented via the [conditionalAssignment](https://github.com/nicklockwood/SwiftFormat/pull/1364) SwiftFormat rule.

#### Reasoning

There are several benefits to using an if/switch expression over simply performing assignment on each branch of the following conditional statement:
1. In most cases, you no longer need to explicitly write a type annotation for the property that is being assigned to.
2. The compiler will diagnose more cases where using a mutable `var` is unnecessary.
3. The resulting syntax is visually ligher because the property name being assigned doesn't need to be written on each branch.

```swift
// BEFORE
// 1. An explicit type annotation is required for the uninitialized property.
// 2. `var` is unnecessary here because `planetLocation` is never modified after being initialized, but the compiler doesn't diagnose.
// 3. The `planetLocation` property name is written on each branch so is redundant and visually noisy
var planetLocation: String
if let star = planet.star {
  planetLocation = "The \(star.name) system"
} else {
  planetLocation = "Rogue planet"
}

print(planetLocation)

// AFTER
// 1. No need to write an explicit `: String` type annotation.
// 2. The compiler correctly diagnoses that the `var` is unnecessary and emits a warning suggesting to use `let` instead. 
// 3. Each conditional branch is simply the value being assigned.
var planetLocation = planet.star {
  "The \(star.name) system"
} else {
  "Rogue planet"
}

print(planetLocation)
```

#### Examples

```swift
// WRONG
let planetLocation: String
if let star = planet.star {
  planetLocation = "The \(star.name) system"
} else {
  planetLocation = "Rogue planet"
}

let planetType: PlanetType
switch planet {
case .mercury, .venus, .earth, .mars:
  planetType = .terrestrial
case .jupiter, .saturn, .uranus, .neptune:
  planetType = .gasGiant
}

let canBeTerraformed: Bool 
if 
  let star = planet.star, 
  !planet.isHabitable,
  planet.isInHabitableZone(of: star) 
{
  canBeTerraformed = true
} else {
  canBeTerraformed = false
}

// RIGHT
let planetLocation = if let star = planet.star {
  "The \(star.name) system"
} else {
  "Rogue planet"
}

let planetType: PlanetType = switch planet {
case .mercury, .venus, .earth, .mars:
  .terrestrial
case .jupiter, .saturn, .uranus, .neptune:
  .gasGiant
}

let canBeTerraformed = if 
  let star = planet.star, 
  !planet.isHabitable,
  planet.isInHabitableZone(of: star) 
{
  true
} else {
  false
}

// ALSO RIGHT. This example cannot be converted to an if/switch expression
// because one of the branches is more than just a single expression.
let planetLocation: String
if let star = planet.star {
  planetLocation = "The \(star.name) system"
} else {
  let actualLocaton = galaxy.name ?? "the universe"
  planetLocation = "Rogue planet somewhere in \(actualLocation)"
}
```


jpsoultanis:
Maybe this is just my bias here, but I'm not in favor of this change. I tend to follow the mantra of "better to be clear than clever". 

In my opinion based on the examples above, the code produced in this new pattern is much harder to understand.

Maybe some developers will prefer this style, and that is ok, but I don't think we should lint away the old approach which has merit.

calda:
@jpsoultanis, interested to hear which aspects of the pattern you find harder to understand. 

--- 

You may have noticed this confusing code in the PR body a few minutes ago: 

```swift
var planetLocation = planet.star {
  "The \(star.name) system"
} else {
  "Rogue planet"
}
```

this was a typo and was not valid Swift, it was supposed to be:

```swift
let planetLocation = if let star = planet.star {
  "The \(star.name) system"
} else {
  "Rogue planet"
}
```

I just updated that example and it is now correct.

jpsoultanis:
> @jpsoultanis, interested to hear which aspects of the pattern you find harder to understand.
> 
> You may have noticed this confusing code in the PR body a few minutes ago:
> 
> ```swift
> var planetLocation = planet.star {
>   "The \(star.name) system"
> } else {
>   "Rogue planet"
> }
> ```
> 
> this was a typo and was not valid Swift, it was supposed to be:
> 
> ```swift
> let planetLocation = if let star = planet.star {
>   "The \(star.name) system"
> } else {
>   "Rogue planet"
> }
> ```
> 
> I just updated that example and it is now correct.

I see, that addresses my own confusion. I still think the "wrong" approach is more explicit and clear, but this may be just because I am used to it.

If I had to articulate it, the issue is that I've trained my brain to scan and when I see an `if let` then I "know" that there's some logic in the block that is executed once the unwrap is made. But with this change, the block actually contains the setting of a var, which contradicts my mental model.

calda:
I think I understand what you mean. I think there is some natural tension with how this formatting defies our existing expectations of how an if / switch statement looks. 

```swift
let planetLocation = if let star = planet.star {
  "The \(star.name) system"
} else {
  "Rogue planet"
}
```

I'm curious to hear your thoughts about this alternative formatting, where we insert a line break before the `if` keyword:

```swift
let planetLocation = 
  if let star = planet.star {
    "The \(star.name) system"
  } else {
    "Rogue planet"
  }
```

This is the formatting used in most of the examples used in the [Swift Evolution proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0380-if-switch-expressions.md) for this feature, and has two improvements over the other formatting:
 1. The `if` statement starts on a new line, which is more naturally compatible with how we're used to scanning / reading if statements (we aren't used to looking for them in the middle of some other line of code).
 2. The conditional branches are always to the right of, and more indented than, the `if` keyword, which is the shape we're used to seeing. (In the other example it seems a little funny how the "true" branch of the code is visually to the left of the condition itself, rather than below / to the right like we're used to).

Curious how your thoughts on this formatting compare to formatting used in the examples in the PR body @jpsoultanis.

Regardless of whether or not we merge this PR, it would make sense to choose a preferred formatting for if / switch expressions. (Even if we don't merge this PR, engineers will be permitted to write code using this functionality).

bachand (commented):
I am glad to see the engagement on this PR!

@calda , I am going to hold off on approving this PR given that there is ongoing discussion. If you feel like we should move ahead with this proposal, please tag me and I can review again with the aim of approving.

I am in favor of this proposal. Related to [this](https://github.com/airbnb/swift/pull/250#issuecomment-1797079165) discussion, though, I would prefer the alternate syntax for the reasons you outlined @calda .

```swift
let planetLocation = 
  if let star = planet.star {
    "The \(star.name) system"
  } else {
    "Rogue planet"
  }
```

I am personally in favor of the "less is more" approach to explicit types in Swift, so I like how this new Swift 5.9 syntax can avoid the need to provide an explicit type in a variable declaration. Related to what @nmkel999 [quoted](https://github.com/airbnb/swift/pull/250#issuecomment-1799278252), I find that not specifying types in code makes it easier to perform sweeping refactors.

I also love that this new syntax causes the compiler to diagnose that `let` can be used instead of `var`.

README.md
```diff
@@ -901,6 +901,111 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
   ```
 
+* <a id='conditional-assignment'></a>(<a href='#conditional-assignment'>link</a>) **When initializing a new property with the result of a conditional statement (e.g. an if or switch statement), use a single if/switch expression where possible** rather than definining an uninitialized property and initializing it on every branch of the following conditional statement. [![SwiftFormat: conditionalAssignment](https://img.shields.io/badge/SwiftFormat-conditionalAssignment-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#conditionalAssignment)
```
bachand, line 904: I wonder if we can have a more specific/descriptive slug than `conditional-assignment` but I don't feel strongly that we need to make a change.

bachand, line 904:
```suggestion
* <a id='conditional-assignment'></a>(<a href='#conditional-assignment'>link</a>) **When initializing a new property with the result of a conditional statement (e.g. an `if` or `switch` statement), use a single if/switch expression where possible** rather than defining an uninitialized property and initializing it on every branch of the following conditional statement. [![SwiftFormat: conditionalAssignment](https://img.shields.io/badge/SwiftFormat-conditionalAssignment-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#conditionalAssignment)
```

README.md
```diff
@@ -901,6 +901,111 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  <details>
+
+  #### Why?
+
+  There are several benefits to using an if/switch expression over simply performing assignment on each branch of the following conditional statement:
+  1. In most cases, you no longer need to explicitly write a type annotation for the property that is being assigned to.
```
bachand, line 911:
```suggestion
  1. In most cases, you no longer need to explicitly write a type annotation for the variable that is being assigned to.
```

My understanding based on the code examples is that the rule only applies to variables. It's possible my terminology is off, though.

I proposed this edit so we can discuss whether "property" or "variable" (or both?) is most the most appropriate term.

jpsoultanis:
@calda I much prefer the syntax you mentioned with the line break:

```swift
let planetLocation = 
  if let star = planet.star {
    "The \(star.name) system"
  } else {
    "Rogue planet"
  }
```

Thanks for your work on this 😄

README.md
bachand: @calda thanks for confirming. I can see how a similar simplification could be made when implementing a property.

```swift
class Foo {
    let anotherNumber = 2
    
    // BEFORE

    var myValue: String {
        let result: String
        if anotherNumber % 2 == 0 {
            result = "even"
        } else {
            result = "odd"
        }
        return result
    }
    
    
    // RIGHT
    
    var myValue: String {
        if anotherNumber % 2 == 0 {
            "even"
        } else {
            "odd"
        }
    }
}
```

Would it make sense for this ^ similar autocorrection to be included in this rule? It's also possible that what I wrote above is distinct from the new Swift 5.9 functionality.

I tried to check if the above code would compile before Swift 5.9 but I am not sure if it's easy to do that now that I've downloaded Xcode 15. I updated my project to build for Swift 4.2 but even after doing that I added this function that uses the new Swift 5.9 conditional assignment feature and the application compiled 🤔 

Screenshot 2023-11-08 at 8 40 30 AM


```swift
  func myFunc() {
    let value = if anotherNumber % 2 == 0 {
      true
    } else {
      false
    }
    print(value)
  }
```

calda: > ```swift
> // BEFORE
> var myValue: String {
>   let result: String
>   if anotherNumber % 2 == 0 {
>     result = "even"
>   } else {
>     result = "odd"
>   }
>   return result
> }
> 
> // AFTER
> var myValue: String {
>   if anotherNumber % 2 == 0 {
>     "even"
>   } else {
>     "odd"
>   }
> }
> ```
>
> Would it make sense for this ^ similar autocorrection to be included in this rule?
    
This is a transformation that could conceivably be implemented as a SwiftFormat rule, but currently isn't. As implemented the `conditionalAssignment` rule adopted in this proposal would convert the initial code to:

```swift
// BEFORE
var myValue: String {
  let result: String = 
    if anotherNumber % 2 == 0 {
      "even"
    } else {
      "odd"
    }
  return result
}
```

You could imagine a separate rule that would simplify away the `let result` / `return result` ceremony and just directly return the RHS expression. This could apply to any time where you define a variable and then return it immediately.

Another note is that if somebody writes this:

```swift
var myValue: String {
  if anotherNumber % 2 == 0 {
    return "even"
  } else {
    return "odd"
  }
}
```

we already automatically update it to:

```swift
var myValue: String {
  if anotherNumber % 2 == 0 {
    "even"
  } else {
    "odd"
  }
}
```

just by applying our existing "omit redundant return" rule. This happened automatically when we updated the SwiftFormat `--swiftversion` to 5.9. See https://github.com/airbnb/swift/pull/249 for a bit more discussion about this.

> I tried to check if the above code would compile before Swift 5.9 but I am not sure if it's easy to do that now that I've downloaded Xcode 15.

I use https://swiftfiddle.com/ to test older versions of the Swift compiler. It supports every version back to 2.2!

The ability to omit the `return` keyword in an if/switch statement like this is a new feature introduced in Swift 5.9.

> I updated my project to build for Swift 4.2 but even after doing that I added this function that uses the new Swift 5.9 conditional assignment feature and the application compiled 🤔

This "language version" setting is distinct from the "compiler version". The language version is a gating mechanism for breaking language changes. For example, our codebase currently has code that compiles in Swift 5.9 but won't compile in Swift 6.0. If you compile our app using the Swift 6.0 compiler, but with the "Swift 5 language mode", the Swift 6.0 compiler will permit all code that was valid under Swift 5. But since we'd be using the Swift 6.0 compiler, we'd also be able to use new features introduced in Swift 6.0.

bachand: > > ```swift
> > // BEFORE
> > var myValue: String {
> >   let result: String
> >   if anotherNumber % 2 == 0 {
> >     result = "even"
> >   } else {
> >     result = "odd"
> >   }
> >   return result
> > }
> > 
> > // AFTER
> > var myValue: String {
> >   if anotherNumber % 2 == 0 {
> >     "even"
> >   } else {
> >     "odd"
> >   }
> > }
> > ```
> >
> > Would it make sense for this ^ similar autocorrection to be included in this rule?
> 
> This is a transformation that could conceivably be implemented as a SwiftFormat rule, but currently isn't. As implemented the `conditionalAssignment` rule adopted in this proposal would convert the initial code to:
> 
> ```swift
> // BEFORE
> var myValue: String {
>   let result: String = 
>     if anotherNumber % 2 == 0 {
>       "even"
>     } else {
>       "odd"
>     }
>   return result
> }
> ```
> 
> You could imagine a separate rule that would simplify away the `let result` / `return result` ceremony and just directly return the RHS expression. This could apply to any time where you define a variable and then return it immediately.

Makes sense! I agree that there's another rule that we could add in the future.

I wanted to discuss this now in case we thought that the best path would be expanding this currently proposed rule. Your observation that we could also condense property implementations that don't leverage conditional assignment, though, has convinced me that we would want a new rule.

> 
> Another note is that if somebody writes this:
> 
> ```swift
> var myValue: String {
>   if anotherNumber % 2 == 0 {
>     return "even"
>   } else {
>     return "odd"
>   }
> }
> ```
> 
> we already automatically update it to:
> 
> ```swift
> var myValue: String {
>   if anotherNumber % 2 == 0 {
>     "even"
>   } else {
>     "odd"
>   }
> }
> ```
> 
> just by applying our existing "omit redundant return" rule. This happened automatically when we updated the SwiftFormat `--swiftversion` to 5.9. See #249 for a bit more discussion about this.

Ah, thanks for mentioning this. I felt like there was a relationship between what I was writing about and omitting redundant return statements. You've made it clear to me!

> > I tried to check if the above code would compile before Swift 5.9 but I am not sure if it's easy to do that now that I've downloaded Xcode 15.
> 
> I use https://swiftfiddle.com/ to test older versions of the Swift compiler. It supports every version back to 2.2!

TIL about https://swiftfiddle.com/. So useful! I see that the simplified property implementation compiles in Swift 5.9 but not in Swift 5.8.

| Swift 5.8 | Swift 5.9 | 
| ---- | ---- |
| Screenshot 2023-11-08 at 2 17 38 PM | Screenshot 2023-11-08 at 2 17 24 PM |

> > I updated my project to build for Swift 4.2 but even after doing that I added this function that uses the new Swift 5.9 conditional assignment feature and the application compiled 🤔
> 
> This "language version" setting is distinct from the "compiler version". The language version is a gating mechanism for breaking language changes. For example, our codebase currently has code that compiles in Swift 5.9 but won't compile in Swift 6.0. If you compile our app using the Swift 6.0 compiler, but with the "Swift 5 language mode", the Swift 6.0 compiler will permit all code that was valid under Swift 5. But since we'd be using the Swift 6.0 compiler, we'd also be able to use new features introduced in Swift 6.0.

👍

calda:
https://github.com/airbnb/swift/pull/253 adopts the formatting we discussed above, where we insert a line break after the assignment operator.

README.md
```diff
@@ -901,6 +901,115 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
   ```
 
+* <a id='conditional-assignment'></a>(<a href='#conditional-assignment'>link</a>) **When initializing a new property with the result of a conditional statement (e.g. an `if` or `switch` statement), use a single if/switch expression where possible** rather than defining an uninitialized property and initializing it on every branch of the following conditional statement. [![SwiftFormat: conditionalAssignment](https://img.shields.io/badge/SwiftFormat-conditionalAssignment-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#conditionalAssignment)
+
+  <details>
+
+  #### Why?
+
+  There are several benefits to using an if/switch expression over simply performing assignment on each branch of the following conditional statement:
+  1. In most cases, you no longer need to explicitly write a type annotation for the variable that is being assigned to.
```
bachand, line 911: I noticed that we add a period after each numbered item here, but we don't have a period in https://github.com/airbnb/swift/pull/253. Let's be consistent, one way or the other.

README.md
```diff
@@ -901,6 +901,115 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  2. The compiler will diagnose more cases where using a mutable `var` is unnecessary.
+  3. The resulting syntax is visually ligher because the property name being assigned doesn't need to be written on each branch.
```
bachand, line 913:
```suggestion
  3. The resulting syntax is visually lighter because the property name being assigned doesn't need to be written on each branch.
```

README.md
```diff
@@ -901,6 +901,115 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  ```swift
+  // BEFORE
+  // 1. An explicit type annotation is required for the uninitialized property.
+  // 2. `var` is unnecessary here because `planetLocation` is never modified after being initialized, but the compiler doesn't diagnose.
+  // 3. The `planetLocation` property name is written on each branch so is redundant and visually noisy
```
bachand, line 919: Let's end these numbered items consistently with a period. @fdiaz taught me that this is a good practice 👍

README.md
```diff
@@ -901,6 +901,115 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    }
+
+  // ALSO RIGHT. This example cannot be converted to an if/switch expression
+  // because one of the branches is more than just a single expression.
+  let planetLocation: String
+  if let star = planet.star {
+    planetLocation = "The \(star.name) system"
+  } else {
+    let actualLocaton = galaxy.name ?? "the universe"
+    planetLocation = "Rogue planet somewhere in \(actualLocation)"
```
bachand, line 1007: 👍

bachand (approved):
I am approving since I believe that there is general alignment on moving ahead with this proposal. Thanks @calda !

Package.swift
```diff
@@ -42,8 +42,8 @@ let package = Package(
 
     .binaryTarget(
       name: "SwiftFormat",
-      url: "https://github.com/calda/SwiftFormat/releases/download/0.53-beta-3/SwiftFormat.artifactbundle.zip",
-      checksum: "ef7283d3dab42809d3faa812dbad71d4f35cd2863db60bc920128fcc12a36a1e"),
+      url: "https://github.com/calda/SwiftFormat/releases/download/0.53-beta-4/SwiftFormat.artifactbundle.zip",
+      checksum: "8506e9f6a434127f9eabe62c0a0349ccfd1e12e7cd7a96ba6a0c8f9d4a596099"),
```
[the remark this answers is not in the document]
calda: Yeah, the build would fail with an invalid checksum

README.md
```diff
@@ -901,6 +901,115 @@ _You can enable the following settings in Xcode by running [this script](resourc
   }
   ```
 
+* <a id='conditional-assignment'></a>(<a href='#conditional-assignment'>link</a>) **When initializing a new property with the result of a conditional statement (e.g. an `if` or `switch` statement), use a single if/switch expression where possible** rather than defining an uninitialized property and initializing it on every branch of the following conditional statement. [![SwiftFormat: conditionalAssignment](https://img.shields.io/badge/SwiftFormat-conditionalAssignment-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#conditionalAssignment)
```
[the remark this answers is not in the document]
calda: Sure, sounds good

PR #253, opened 2023-11-17
Add rule to add a line break after the assignment operator before a multi-line `if` or `switch` expressions

calda:
#### Summary

This PR proposes a new rule to add a line break after the assignment operator before a multi-line `if` or `switch` expression.

Autocorrection for this rule is implemented in the `wrapMultilineConditionalAssignment` SwiftFormat rule, added in https://github.com/nicklockwood/SwiftFormat/pull/1574.

#### Reasoning

This makes it so that `if` and `switch` expressions always have the same "shape" as standard `if` and `switch` statements, where:
1. The `if` / `switch` keyword is always the left-most token on a dedicated line of code
2. The conditional branches are always to the right of and below the `if` / `switch` keyword.

This is most consistent with the formatting of `if` / `switch` statements in other contexts, and thus makes it easier to recognize that the code is using an `if` or `switch` expression at a glance.

This is independent of #250 and could be accepted even if we reject #250, but #250 depends on these changes given the discussion [here](https://github.com/airbnb/swift/pull/250#issuecomment-1797079165).

#### Examples

```swift
// WRONG. Should have a line break after the `=`. 
let planetLocation = if let star = planet.star {
  "The \(star.name) system"
 } else {
  "Rogue planet"
}

// WRONG. The `=` should be on the line before the `if` / `switch` expression. 
let planetLocation 
  = if let star = planet.star {
    "The \(star.name) system"
  } else {
    "Rogue planet"
  }
  
// RIGHT 
let planetLocation = 
  if let star = planet.star {
    "The \(star.name) system"
  } else {
    "Rogue planet"
  }
  
// RIGHT 
let planetLocation = 
  if let star = planet.star {
    "The \(star.name) system"
  } else {
    "Rogue planet"
  }
  
// RIGHT
let planetType: PlanetType =
  switch planet {
  case .mercury, .venus, .earth, .mars:
    .terrestrial
  case .jupiter, .saturn, .uranus, .neptune:
    .gasGiant
  }
  
// ALSO RIGHT. A line break is not required because the declaration fits on a single line. 
let planet = if useFirst { mercury } else { neptune }

// ALSO RIGHT. A line break is always permitted if it helps with readability.
let planet =
  if useFirst { mercury } else { neptune }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
     …
   }
   ```
+  
+  </details>
+
+* <a id='wrap-multiline-conditional-assignment'></a>(<a href='#wrap-multiline-conditional-assignment'>link</a>) **Add a line break after the assignment operator (`=`) before a multi-line `if` or `switch` expression**, and indent the following `if` / `switch` expression. If the declaration fits on a single line, a line break not required. [![SwiftFormat: wrapMultilineConditionalAssignment](https://img.shields.io/badge/SwiftFormat-wrapMultilineConditionalAssignment-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapMultilineConditionalAssignment)
+
+  <details>
+
+  #### Why?
+  
+  This makes it so that `if` and `switch` expressions always have the same "shape" as standard `if` and `switch` statements, where:
+  1. The `if` / `switch` keyword is always the left-most token on a dedicated line of code
+  2. The conditional branches are always to the right of and below the `if` / `switch` keyword
+
+  This is most consistent with the formatting of `if` / `switch` statements in other contexts, and thus makes it easier to recognize that the code is using an `if` or `switch` expression at a glance. 
+  
+  ```swift
+  // WRONG. Should have a line break after the `=`. 
+  let planetLocation = if let star = planet.star {
+    "The \(star.name) system"
+   } else {
+    "Rogue planet"
+  }
+
+  // WRONG. The `=` should be on the line before the `if` / `switch` expression. 
+  let planetLocation 
+    = if let star = planet.star {
+      "The \(star.name) system"
+    } else {
+      "Rogue planet"
+    }
+
+  // WRONG. `if` expression should be indented.
+  let planetLocation =
+  if let star = planet.star {
+    "The \(star.name) system"
+  } else {
+    "Rogue planet"
+  }
+    
+  // RIGHT 
+  let planetLocation = 
+    if let star = planet.star {
+      "The \(star.name) system"
+    } else {
+      "Rogue planet"
+    }
+    
+  // RIGHT 
+  let planetLocation = 
+    if let star = planet.star {
+      "The \(star.name) system"
+    } else {
+      "Rogue planet"
+    }
+    
+  // RIGHT
+  let planetType: PlanetType =
+    switch planet {
+    case .mercury, .venus, .earth, .mars:
+      .terrestrial
+    case .jupiter, .saturn, .uranus, .neptune:
+      .gasGiant
+    }
+    
+  // ALSO RIGHT. A line break is not required because the declaration fits on a single line. 
+  let planet = if useFirst { mercury } else { neptune }
+
+  // ALSO RIGHT. A line break is permitted if it helps with readability.
+  let planet =
+    if useFirst { mercury } else { neptune }
+  ```
+  
+  </details>
 
 * <a id='wrap-guard-else'></a>(<a href='#wrap-guard-else'>link</a>) **Add a line break before the `else` keyword in a multi-line guard statement.** For single-line guard statements, keep the `else` keyword on the same line as the `guard` keyword. The open brace should immediately follow the `else` keyword. [![SwiftFormat: elseOnSameLine](https://img.shields.io/badge/SwiftFormat-elseOnSameLine-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#elseOnSameLine)
 
--- a/README.md
+++ b/README.md
@@ -903,7 +903,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   
   </details>
 
-* <a id='wrap-multiline-conditional-assignment'></a>(<a href='#wrap-multiline-conditional-assignment'>link</a>) **Add a line break after the assignment operator (`=`) before a multi-line `if` or `switch` expression**, and indent the following `if` / `switch` expression. If the declaration fits on a single line, a line break not required. [![SwiftFormat: wrapMultilineConditionalAssignment](https://img.shields.io/badge/SwiftFormat-wrapMultilineConditionalAssignment-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapMultilineConditionalAssignment)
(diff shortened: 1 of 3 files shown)


Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -91,3 +91,4 @@
 --rules elseOnSameLine
 --rules sortTypealiases
 --rules preferForLoop
+--rules wrapMultilineConditionalAssignment
```
calda, line 94: After https://github.com/nicklockwood/SwiftFormat/pull/1574 is merged, I'll pull in an updated version of SwiftFormat that actually includes this rule implementation.

README.md
```diff
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
     …
   }
   ```
+  
+  </details>
```
bachand, line 904: Nice catch!

README.md
```diff
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  <details>
+
+  #### Why?
+  
+  This makes it so that `if` and `switch` expressions always have the same "shape" as standard `if` and `switch` statements, where:
+  1. The `if` / `switch` keyword is always the left-most token on a dedicated line of code
+  2. The conditional branches are always to the right of and below the `if` / `switch` keyword
+
+  This is most consistent with the formatting of `if` / `switch` statements in other contexts, and thus makes it easier to recognize that the code is using an `if` or `switch` expression at a glance.
```
bachand, line 916:
```suggestion
  This is most consistent with how the `if` / `switch` keywords are used for control flow, and thus makes it easier to recognize that the code is using an `if` or `switch` expression at a glance. 
```
I'm not 100% sure on this but I wanted to throw out the idea.

README.md
```diff
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  #### Why?
+  
+  This makes it so that `if` and `switch` expressions always have the same "shape" as standard `if` and `switch` statements, where:
+  1. The `if` / `switch` keyword is always the left-most token on a dedicated line of code
+  2. The conditional branches are always to the right of and below the `if` / `switch` keyword
+
+  This is most consistent with the formatting of `if` / `switch` statements in other contexts, and thus makes it easier to recognize that the code is using an `if` or `switch` expression at a glance. 
+  
+  ```swift
+  // WRONG. Should have a line break after the `=`.
```
bachand, line 919:
```suggestion
  // WRONG. Should have a line break after the first `=`. 
```

README.md
```diff
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  
+  ```swift
+  // WRONG. Should have a line break after the `=`. 
+  let planetLocation = if let star = planet.star {
+    "The \(star.name) system"
+   } else {
+    "Rogue planet"
+  }
+
+  // WRONG. The `=` should be on the line before the `if` / `switch` expression.
```
bachand, line 926:
```suggestion
  // WRONG. The first `=` should be on the line of the variable being assigned.
```

This is a more affirmative way of saying the same thing. I think we should include the "first" but I don't feel strongly about the rest of the change.

README.md
```diff
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+      "Rogue planet"
+    }
+    
+  // RIGHT 
+  let planetLocation = 
+    if let star = planet.star {
+      "The \(star.name) system"
+    } else {
+      "Rogue planet"
+    }
```
bachand, line 956: Did we mean to repeat this example? At first I wondered if it was for symmetry with the WRONG examples, but I see this code snippet listed 3 times in WRONG and 2 times in RIGHT, so I'm unsure now.

README.md
```diff
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    
+  // RIGHT
```
bachand, line 958: Should we have a WRONG example for a `switch`?

README.md
```diff
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    case .jupiter, .saturn, .uranus, .neptune:
+      .gasGiant
+    }
+    
+  // ALSO RIGHT. A line break is not required because the declaration fits on a single line. 
+  let planet = if useFirst { mercury } else { neptune }
+
+  // ALSO RIGHT. A line break is permitted if it helps with readability.
+  let planet =
+    if useFirst { mercury } else { neptune }
```
bachand, line 972: Would this be valid?

```swift
let moonName = if let moon = moon { moon.name } else { "none" }
```

It's a bit odd to see two `=` on the same line... I think it's too much of a nuance to create a rule around it, but I wonder if these last two examples would be more powerful if we covered the double `=` situation instead.

```suggestion
  // ALSO RIGHT. A line break is not required because the declaration fits on a single line.
  let moonString = if let moon = moon { moon.name } else { "none" }

  // ALSO RIGHT. A line break is permitted if it helps with readability.
  let moonString =
    if let moon = moon { moon.name } else { "none" }
```

The double `=` situation seems like a great time to add a line break even if it's not strictly required.

README.md
```diff
@@ -900,6 +900,79 @@ _You can enable the following settings in Xcode by running [this script](resourc
     …
   }
   ```
+  
+  </details>
+
+* <a id='wrap-multiline-conditional-assignment'></a>(<a href='#wrap-multiline-conditional-assignment'>link</a>) **Add a line break after the assignment operator (`=`) before a multi-line `if` or `switch` expression**, and indent the following `if` / `switch` expression. If the declaration fits on a single line, a line break not required. [![SwiftFormat: wrapMultilineConditionalAssignment](https://img.shields.io/badge/SwiftFormat-wrapMultilineConditionalAssignment-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapMultilineConditionalAssignment)
```
bachand, line 906:
```suggestion
* <a id='wrap-multiline-conditional-assignment'></a>(<a href='#wrap-multiline-conditional-assignment'>link</a>) **Add a line break after the assignment operator (`=`) before a multi-line `if` or `switch` expression**, and indent the following `if` / `switch` expression. If the declaration fits on a single line, a line break is not required. [![SwiftFormat: wrapMultilineConditionalAssignment](https://img.shields.io/badge/SwiftFormat-wrapMultilineConditionalAssignment-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#wrapMultilineConditionalAssignment)
```

bachand (approved):
Love it! Thanks @calda .

README.md
calda: Good idea

README.md
calda: This was a mistake, thanks!

calda, line 956:
```suggestion
```

README.md
calda: Sure

README.md
calda: > Would this be valid?
>
> `let moonName = if let moon = moon { moon.name } else { "none" }`

Yeah, that example is perfectly valid. I like the idea of using an `=` in the example. I'll just change it to something other than `if let moon = moon {`, which our style guide would collapse to `if let moon {`.

README.md
```diff
@@ -957,11 +958,11 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
     
   // ALSO RIGHT. A line break is not required because the declaration fits on a single line. 
-  let planet = if useFirst { mercury } else { neptune }
+  let moonName = if let moon = planet.moon { moon.name } else { "none" }
 
   // ALSO RIGHT. A line break is permitted if it helps with readability.
-  let planet =
-    if useFirst { mercury } else { neptune }
+  let moonName =
+    if let moon = planet.moon { moon.name } else { "none" }
```
bachand, line 965: Nice!

PR #263, opened 2024-03-13
Add rule to infer property types from the right-hand-side value rather than writing the type explicitly on the left-hand side

calda:
_Please react with 👍/👎 below if you agree or disagree with this proposal._

#### Summary

This PR proposes a new rule to prefer letting the type of a property be inferred from the right-hand-side value rather than writing the type explicitly on the left-hand side:

```swift
// WRONG
let sun: Star = .init(mass: 1.989e30)
let earth: Planet = .earth

// RIGHT
let sun = Star(mass: 1.989e30)
let earth = Planet.earth
```

Autocorrect support for this rule is implemented in https://github.com/nicklockwood/SwiftFormat/pull/1640.

#### Reasoning

We already have a rule to omit redundant types. It converts `let sun: Star = Star()` to `let sun = Star()`, rather than to `let sun: Star = .init()`.

The syntax with inferred types is more idiomatic, and is much more common, than the syntax with explicit types.

I briefly checked the frequency of these two styles in our codebase with a regex search:
 - Syntax with implicit types (`let sun = Star()` or `let earth = Planet.earth`, etc):
   - ~20,000 matches using the regex `((let)|(var)) \w+ = [A-Z]+\w+(\.|\()`
 - Syntax with explicit types (`let sun: Star = .init()`, etc)
   - ~5,000 matches using the regex `((let)|(var)) \w+: \w+ = \.`.

This rule only applies to simple cases where the right-hand side starts with a leading dot (meaning we know that the RHS value refers to a static member of the type written on the LHS). For other cases, we don't state a preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,7 +388,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   // ALSO RIGHT: Explicit types can be necessary when the right-hand side has
   // a different type from the one written explicitly on the left-hand side.
-  let nautralSatellite: PlanetaryBody? = Moon(mass: 7.347e22)
+  let naturalSatellite: PlanetaryBody? = Moon(mass: 7.347e22)
   let moon: PlanetaryBody? = nil // nil literals are typeless
   let numberOfPlanets: UInt = 8 // integer literals default to `Int`
   let sunMass: CGFloat = 1.989e30 // floating-point literals default to `Double`
@@ -399,6 +399,40 @@ _You can enable the following settings in Xcode by running [this script](resourc
   let numberOfPlanets = UInt(8)
   let sunMass = CGFloat(1.989e30)
   let planets = [Planet]()
+  ```
+  
+  There are some rarer cases where the inferred type syntax has a different meaning than the explicit type syntax. In these cases, the explicit type syntax is still permitted:
+  
+  ```swift
+  extension String {
+    static let earth = "Earth"
+  }
+
+  // RIGHT: If the property's type is optional, moving the optional type 
+  // to the right-hand side may result in invalid code.  
+  let planetName: String? = .earth
+  
+  // WRONG: fails with "error: type 'String?' has no member 'foo'"
+  let planetName = String?.earth
+  ```
+  
+  ```swift
+  struct SaturnOutline: ShapeStyle { ... }
+  
+  extension ShapeStyle where Self == SaturnOutline {
+    static var saturnOutline: SaturnOutline { 
+      SaturnOutline() 
+    }
+  }
+  
+  // RIGHT: If the property's type is an existential / protocol type, moving the type
+  // to the right-hand side will result in invalid code if the value is defined in an
+  // extension like `extension ShapeStyle where Self == SaturnOutline`.
+  // SwiftFormat autocorrect detects this case by checking for the existential `any` keyword.
+  let myShape1: any ShapeStyle = .saturnOutline
+  
+  // WRONG: fails with "error: static member 'saturnOutline' cannot be used on protocol metatype '(any ShapeStyle).Type'"
+  let myShape2 = (any ShapeStyle).myShape
   ```
 
   </details>
--- a/README.md
+++ b/README.md
@@ -333,25 +333,72 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   ```swift
   // WRONG
-  let host: Host = Host()
+  let sun: Star = Star(mass: 1.989e30)
+  let earth: Planet = Planet.earth
 
   // RIGHT
-  let host = Host()
+  let sun = Star(mass: 1.989e30)
+  let earth = Planet.earth
+
+  // WRONG: Most literals provide a default type that can be inferred.
+  let enableGravity: Bool = true
+  let numberOfPlanets: Int = 8
+  let sunMass: Double = 1.989e30
+
+  // RIGHT
+  let enableGravity = true
+  let numberOfPlanets = 8
+  let sunMass = 1.989e30
+  
+  // WRONG: Types can be inferred from if/switch expressions as well if each branch has the same explicit type.
+  let smallestPlanet: Planet =
+    if treatPlutoAsPlanet {
+      Planet.pluto
+    } else {
+      Planet.mercury
+    }
+
+  // RIGHT
+  let smallestPlanet =
+    if treatPlutoAsPlanet {
+      Planet.pluto
+    } else {
+      Planet.mercury
+    }
   ```
 
+  </details>
+
+* <a id='infer-property-types'></a>(<a href='#infer-property-types'>link</a>) **Prefer letting the type of a property be inferred from the right-hand-side value rather than writing the type explicitly on the left-hand side.** [![SwiftFormat: preferInferredTypes](https://img.shields.io/badge/SwiftFormat-preferInferredTypes-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#preferInferredTypes)
+
+  <details>
+
   ```swift
-  enum Direction {
-    case left
-    case right
-  }
+  // WRONG
+  let sun: Star = .init(mass: 1.989e30)
+  let earth: Planet = .earth
 
-  func someDirection() -> Direction {
-    // WRONG
-    return Direction.left
+  // RIGHT
+  let sun = Star(mass: 1.989e30)
+  let earth = Planet.earth
 
-    // RIGHT
-    return .left
-  }
+  // ALSO RIGHT: Explicit types are required when there is no right-hand-side value.
+  let sun: Star
+  let earth: Planet
+
+  // ALSO RIGHT: Explicit types can be necessary when the right-hand side has
+  // a different type from the one written explicitly on the left-hand side.
+  let nautralSatellite: PlanetaryBody? = Moon(mass: 7.347e22)
+  let moon: PlanetaryBody? = nil // nil literals are typeless
+  let numberOfPlanets: UInt = 8 // integer literals default to `Int`
+  let sunMass: CGFloat = 1.989e30 // floating-point literals default to `Double`
+  let planets: [Planet] = [] // empty collection literals are typeless
+
+  // ALSO RIGHT: Some of the examples above can also be written idiomatically without an explicit type on
+  // the left-hand-side, by instead giving the right-hand side value an explicit type. Either style is fine.
+  let numberOfPlanets = UInt(8)
+  let sunMass = CGFloat(1.989e30)
+  let planets = [Planet]()
   ```
 
   </details>
--- a/README.md
+++ b/README.md
@@ -340,6 +340,13 @@ _You can enable the following settings in Xcode by running [this script](resourc
   let sun = Star(mass: 1.989e30)
   let earth = Planet.earth
 
+  // NOT RECOMMENDED. However, since the linter doesn't have full type information, this is not enforced automatically.
+  let moon: Moon = earth.moon // returns `Moon`
+
+  // RIGHT
+  let moon = earth.moon
+  let moon: PlanetaryBody? = earth.moon
+
   // WRONG: Most literals provide a default type that can be inferred.
   let enableGravity: Bool = true
   let numberOfPlanets: Int = 8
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -333,25 +333,72 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
 
-  func someDirection() -> Direction {
-    // WRONG
-    return Direction.left
+  // RIGHT
+  let sun = Star(mass: 1.989e30)
+  let earth = Planet.earth
 
-    // RIGHT
-    return .left
```
calda, line 353: This is a nice example and is good advice in general, but doesn't reflect the behavior of the SwiftFormat `redundantType` rule that we use.

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -97,3 +97,4 @@
 --rules preferForLoop
 --rules conditionalAssignment
 --rules wrapMultilineConditionalAssignment
+--rules preferInferredTypes
```
calda, line 100: Will pull in a SwiftFormat build with this new rule after https://github.com/nicklockwood/SwiftFormat/pull/1640 is merged.

jqsilver (approved):

README.md
```diff
@@ -333,25 +333,106 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
   }
+  
+  // RIGHT: If the property's type is an existential / protocol type, moving the type
+  // to the right-hand side will result in invalid code if the value is defined in an
+  // extension like `extension ShapeStyle where Self == SaturnOutline`.
+  // SwiftFormat autocorrect detects this case by checking for the existential `any` keyword.
+  let myShape1: any ShapeStyle = .saturnOutline
+  
+  // WRONG: fails with "error: static member 'saturnOutline' cannot be used on protocol metatype '(any ShapeStyle).Type'"
+  let myShape2 = (any ShapeStyle).myShape
```
calda, line 435: By the way -- for this we have the ability to tell SwiftFormat a list of symbols to exclude from this rule (`--preservesymbols ShapeStyle,OtherType,etc). Inside the apps repo we could exclude any type or property name where we frequently use this syntax. Two relevant examples are:
 - `UnresolvedShapeStyle` 
   - all existing callsites use `any UnresolvedShapeStyle` so this is already handled properly
 - `canonicalFake`
   - there are about a dozen examples of `let experimentsService: ExperimentsService = .canonicalFake`. It seems like a good idea to exclude this using `--preservesymbols canonicalFake`. Updating the callsites to `let experimentsService: any ExperimentsService = .canonicalFake` also works.

README.md
```diff
@@ -333,25 +333,72 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   ```swift
   // WRONG
-  let host: Host = Host()
+  let sun: Star = Star(mass: 1.989e30)
```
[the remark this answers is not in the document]
bachand: Thank you!

README.md
```diff
@@ -333,25 +333,106 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  let earth: Planet = Planet.earth
 
   // RIGHT
-  let host = Host()
+  let sun = Star(mass: 1.989e30)
+  let earth = Planet.earth
+
+  // WRONG: Most literals provide a default type that can be inferred.
+  let enableGravity: Bool = true
+  let numberOfPlanets: Int = 8
```
bachand, line 345: @calda how do we handle/guide on the case where you want to tell the language that a literal should be interpreted in a special way? For example, one way to get a floating-point value of 100 is to write

```swift
let myFloatingPointValue: Double = 100
```

which will produce a different result than

```swift
let myFloatingPointValue = 100 // implicitly of type `Int`
```

In this example, another option is to write
```swift
let myFloatingPointValue = 100.0 // implicitly of type `Double`
```

but even then it's necessary to write an explicit type in some cases, like if you explicitly wanted a `Float` instead of `Double`
```swift
let myFloatingPointValue: Float = 100
```

bachand: I am curious how our autocorrection handles this case where removing the type can change the meaning of the code, and I also think that situations like this may warrant some discussion in the prose of this rule.

README.md
```diff
@@ -333,25 +333,106 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    if treatPlutoAsPlanet {
+      Planet.pluto
+    } else {
+      Planet.mercury
+    }
   ```
 
+  </details>
+
+* <a id='infer-property-types'></a>(<a href='#infer-property-types'>link</a>) **Prefer letting the type of a property be inferred from the right-hand-side value rather than writing the type explicitly on the left-hand side.** [![SwiftFormat: preferInferredTypes](https://img.shields.io/badge/SwiftFormat-preferInferredTypes-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#preferInferredTypes)
```
bachand, line 372: I see that our rule says "property types". Do we need to say "variable or property types"? In the code

```swift
func printSunDetails() {
  let sun = Star(mass: 1.989e3)
  print(sun)
}
```

I consider `sun` a variable not a property. It's also possible that there is "property" has a special meaning that I am not considering in the context of the grammar of the language.

README.md
bachand: In my testing I see that the linter is smart enough to correct `let numberOfPlanets: Int = 8` to `let numberOfPlanets = 8` but not correct `let numberOfPlanets: Float = 8`. Wow! I wonder if you happen to have a code pointer to how this is implemented.

README.md
```diff
@@ -333,25 +333,106 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  // ALSO RIGHT: Explicit types are required when there is no right-hand-side value.
+  let sun: Star
+  let earth: Planet
+
+  // ALSO RIGHT: Explicit types can be necessary when the right-hand side has
+  // a different type from the one written explicitly on the left-hand side.
+  let naturalSatellite: PlanetaryBody? = Moon(mass: 7.347e22)
+  let moon: PlanetaryBody? = nil // nil literals are typeless
+  let numberOfPlanets: UInt = 8 // integer literals default to `Int`
+  let sunMass: CGFloat = 1.989e30 // floating-point literals default to `Double`
```
bachand, line 394: This one is a bit nuanced as I recall that `CGFloat` is a typealias to `Double` on 64-bit platforms... but it does differ for 32-bit platforms.

Screenshot 2024-04-04 at 6 34 10 PM

I recommend that we change the example to the following to avoid anyone needing to consider these ^ nuances that aren't important to understand this rule.
```suggestion
  let sunMass: Float = 1.989e30 // floating-point literals default to `Double`
```

README.md
```diff
@@ -333,25 +333,106 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  let planets: [Planet] = [] // empty collection literals are typeless
+
+  // ALSO RIGHT: Some of the examples above can also be written idiomatically without an explicit type on
+  // the left-hand-side, by instead giving the right-hand side value an explicit type. Either style is fine.
+  let numberOfPlanets = UInt(8)
```
bachand, line 399: I want to call out that I think in theory `let numberOfPlanets = UInt(8)` and `let numberOfPlanets: UInt = 8` could be subtly different. My understanding is that the latter will invoke the initializer from [the `ExpressibleByIntegerLiteral ` conformance](https://developer.apple.com/documentation/swift/expressiblebyintegerliteral) while the former could invoke an explicit initializer on the type. However, I am having trouble coming up with code where this would be problematic, so maybe it's not a real issue in practice.

README.md
bachand: Thanks for flagging this @calda . I was wondering about this... My recollection is that eventually it's going to be required that we use `any` for any existential/protocol types but that this requirement is being slowly rolled out. Is your understanding also that eventually we will always need to prefix an existential/protocol type with `any`? If so, I think it's less problematic if we need to work around certain situations like the above ones you highlighted for the time being.

bachand: I am trying to understand if there would be cases when we still need `--preservesymbols` if/when the language requires that all existential/protocol types are preceded by `any`.

bachand (commented):
@calda this is looking great! I was hoping to talk through some comments before approving if that's OK but I feel like we are very close 👍

README.md
bachand: Btw I also saw as I continued to review that this is covered in more detail in the next rule, though I wonder if we should move some of the content from the next rule up to this one since this rule is the one focused on removing types.

calda: I implemented this back in https://github.com/nicklockwood/SwiftFormat/pull/921! Here's the code: https://github.com/nicklockwood/SwiftFormat/pull/921/files#diff-1f5cbb52e333eb61c8d6fef2bd765e3e8e6fa8f3101f91f98a9dfce9e1ef480dR1363

README.md
calda: Ah, yes, let's not use `CGFloat` as an example. Good thinking.

README.md
calda: Maybe we remove this example. I've basically never seen people write this type of code so maybe it's not really worth mentioning. Sometimes less is more!

README.md
calda: I saw on the Swift Evolution forums that they decided to not make `any` mandatory in Swift 6 mode after all. Although after a quick search on the forums I can't actually find where they said that. 

This doesn't necessarily confirm it, but if you build code like [this](https://swiftfiddle.com/jazi4p4ji5eftlgxzvigkq2wi4) using a nightly Swift 6 build it doesn't require `any`.

> I am trying to understand if there would be cases when we still need --preservesymbols if/when the language requires that all existential/protocol types are preceded by any.

If all call sites used the existential `any` syntax, we wouldn't need to use `--preservesymbols`

README.md
calda: You're right, I think "variable" is more correct for this examples. Maybe we should say "type of a variable or property" since the rule applies to both variables and properties on types.

README.md
calda: > I also saw as I continued to review that this is covered in more detail in the next rule, though I wonder if we should move some of the content from the next rule up to this one since this rule is the one focused on removing types.

Good idea. I think the next rule probably goes in to too much detail about literals, and that content would be a better fit for the discussion about the `redundantType` rule.

README.md
calda: Found the quote: https://forums.swift.org/t/progress-toward-the-swift-6-language-mode/68315

> The Language Steering Group has decided that one previously-accepted upcoming feature, ExistentialAny, will not be enabled by default in Swift 6. [SE-0335: Introduce existential any](https://github.com/apple/swift-evolution/blob/main/proposals/0335-existential-any.md) introduces the any keyword to identify existential types. The proposal also specifies that "bare" protocol names will no longer be permitted as types---they must either use any or some, as appropriate---under the upcoming feature flag ExistentialAny. Given the concerns about migration to consistent use of existential any in the language, and the expectation that additional language improvements will be coming that might affect the end result of that migration, the Language Steering Group is deferring the source-incompatible changes in SE-0335 to a future language revision.

It sounds like they didn't rule out this direction, but rather just deferred it for longer (probably a very long time, if we have to wait for Swift 7!)

README.md
calda: I added an asana task to our internal board for us to look at adding autocorrect for this sort of case

README.md
```diff
@@ -333,25 +333,125 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  }
+
+  // WRONG: fails with "error: static member 'saturnOutline' cannot be used on protocol metatype '(any ShapeStyle).Type'"
+  let myShape2 = (any ShapeStyle).myShape
+
+  // RIGHT: If the property's type is an existential / protocol type, moving the type
+  // to the right-hand side will result in invalid code if the value is defined in an
+  // extension like `extension ShapeStyle where Self == SaturnOutline`.
+  // SwiftFormat autocorrect detects this case by checking for the existential `any` keyword.
+  let myShape1: any ShapeStyle = .saturnOutline
```
calda, line 454: I spent a bit of time thinking about this case in more detail, that `let myShape1: ShapeStyle = .saturnOutline` is allowed but `let myShape1: = ShapeStyle.saturnOutline` isn't. I couldn't understand why this wasn't allowed.

This syntax was introduced in [SE-0299](https://github.com/apple/swift-evolution/blob/main/proposals/0299-extend-generic-static-member-lookup.md). This [came up briefly](https://forums.swift.org/t/se-0299-extending-static-member-lookup-in-generic-contexts/43958/33) in the review thread, and Pavel mentioned that there was more discussion in the pitch thread. A found the most compelling discussion [here](https://forums.swift.org/t/proposal-static-member-lookup-on-protocol-metatypes/41946/48) in the pitch thread.

They intentionally don't frame this as adding a static member to the protocol type, but rather a shorthand syntax that lets you omit the name of the concrete type.

Basically, the thinking is that this code:

```swift
let myShape1: ShapeStyle = .saturnOutline
```

is actually implicit shorthand for _this code_:

```swift
let myShape1: ShapeStyle = SaturnOutline.saturnOutline
```

_not_ shorthand for this code:

```swift
let myShape1: ShapeStyle = ShapeStyle.saturnOutline
```

This is subtle but distinct, and is a reasonable explanation for why `ShapeStyle.saturnOutline` isn't allowed.

README.md
calda: In the apps repo PR that I merged today, I included `--preservesymbols canonicalFake`. This is a common and reasonable pattern that I think we should continue supporting. It's not common to use the `any ExperimentService` spelling today, so it felt too weird to require it in this specific case due to this limitation of the linter.

README.md
calda: This problem should also away _eventually_, once `any` is required in a future language mode (just [not Swift 6](https://github.com/airbnb/swift/pull/263#discussion_r1561395487)).

PR #268, opened 2024-04-16
Return a non-zero exit code when SwiftFormat autocorrect applies code changes

calda:
This PR updates the SPM package plugin to return a non-zero exit code when SwiftFormat autocorrect applies code changes, using the new `--strict` flag added in https://github.com/nicklockwood/SwiftFormat/pull/1676.

This matches the behavior of our internal linting tool.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -42,8 +42,8 @@ let package = Package(
 
     .binaryTarget(
       name: "SwiftFormat",
-      url: "https://github.com/calda/SwiftFormat/releases/download/0.54-beta-5/SwiftFormat.artifactbundle.zip",
-      checksum: "7447986db45a51164d23672c07f971406a4c0589b0c423fcb85e95ed8f8e7e48"),
+      url: "https://github.com/calda/SwiftFormat/releases/download/0.54-beta-6/SwiftFormat.artifactbundle.zip",
+      checksum: "2b7213c898f0914713b5d8a75e4ed7c560552f3a9e69f31dbcd38c99b878d738"),
 
     .binaryTarget(
       name: "SwiftLintBinary",
--- a/Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
+++ b/Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
@@ -19,15 +19,15 @@ final class AirbnbSwiftFormatToolTest: XCTestCase {
 
     let error = runFormatTool(
       with: MockCommands(
-        swiftFormat: {
+        swiftFormat: { _ in
           ranSwiftFormat = true
           return EXIT_SUCCESS
         },
-        swiftLint: {
+        swiftLint: { _ in
           ranSwiftLint = true
           return EXIT_SUCCESS
         },
-        swiftLintAutocorrect: {
+        swiftLintAutocorrect: { _ in
           ranSwiftLintAutocorrect = true
           return EXIT_SUCCESS
         }))
@@ -46,15 +46,15 @@ final class AirbnbSwiftFormatToolTest: XCTestCase {
     let error = runFormatTool(
       arguments: ["--lint"],
       with: MockCommands(
-        swiftFormat: {
+        swiftFormat: { _ in
           ranSwiftFormat = true
           return EXIT_SUCCESS
         },
-        swiftLint: {
+        swiftLint: { _ in
           ranSwiftLint = true
           return EXIT_SUCCESS
         },
-        swiftLintAutocorrect: {
+        swiftLintAutocorrect: { _ in
           ranSwiftLintAutocorrect = true
           return EXIT_SUCCESS
         }))
@@ -74,18 +74,22 @@ final class AirbnbSwiftFormatToolTest: XCTestCase {
 
     let error = runFormatTool(
       with: MockCommands(
-        swiftFormat: {
+        swiftFormat: { command in
           ranSwiftFormat = true
 
-          // When autocorrecting SwiftFormat returns EXIT_SUCCESS
-          // even if there were violations that were fixed
-          return EXIT_SUCCESS
+          // When autocorrecting, SwiftFormat returns EXIT_SUCCESS even if there were
+          // violations that were fixed, unless you pass in the --strict option.
+          if command.arguments.contains("--strict") {
+            return SwiftFormatExitCode.lintFailure
+          } else {
+            return EXIT_SUCCESS
+          }
         },
-        swiftLint: {
+        swiftLint: { _ in
           ranSwiftLint = true
           return SwiftLintExitCode.lintFailure
         },
-        swiftLintAutocorrect: {
+        swiftLintAutocorrect: { _ in
           ranSwiftLintAutocorrect = true
 
           // When autocorrecting SwiftLint returns EXIT_SUCCESS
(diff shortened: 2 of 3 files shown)


Package.swift
```diff
@@ -42,8 +42,8 @@ let package = Package(
 
     .binaryTarget(
       name: "SwiftFormat",
-      url: "https://github.com/calda/SwiftFormat/releases/download/0.54-beta-5/SwiftFormat.artifactbundle.zip",
-      checksum: "7447986db45a51164d23672c07f971406a4c0589b0c423fcb85e95ed8f8e7e48"),
+      url: "https://github.com/calda/SwiftFormat/releases/download/0.54-beta-6/SwiftFormat.artifactbundle.zip",
```
bachand, line 45: I checked that [this release](https://github.com/calda/SwiftFormat/releases/tag/0.54-beta-6) includes the `--strict` change.

Package.swift
```diff
@@ -42,8 +42,8 @@ let package = Package(
...
+      checksum: "2b7213c898f0914713b5d8a75e4ed7c560552f3a9e69f31dbcd38c99b878d738"),
```
bachand, line 46: @calda how do you find/compute this value? Asking to learn.

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
```diff
@@ -74,26 +74,63 @@ final class AirbnbSwiftFormatToolTest: XCTestCase {
...
+        swiftFormat: { command in
           ranSwiftFormat = true
 
-          // When autocorrecting SwiftFormat returns EXIT_SUCCESS
-          // even if there were violations that were fixed
-          return EXIT_SUCCESS
+          // When autocorrecting, SwiftFormat returns EXIT_SUCCESS even if there were
+          // violations that were fixed, unless you pass in the --strict option.
+          if command.arguments.contains("--strict") {
+            return SwiftFormatExitCode.lintFailure
```
bachand, line 83: 👍

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
```diff
@@ -74,26 +74,63 @@ final class AirbnbSwiftFormatToolTest: XCTestCase {
...
+        swiftLintAutocorrect: { _ in
           ranSwiftLintAutocorrect = true
 
           // When autocorrecting SwiftLint returns EXIT_SUCCESS
           // even if there were violations that were fixed
           return EXIT_SUCCESS
         }))
 
-    XCTAssertEqual(error as? ExitCode, ExitCode(SwiftFormatExitCode.lintFailure))
+    XCTAssertEqual(error as? ExitCode, .failure)
```
bachand, line 100: Is this the definition of `ExitCode`? https://apple.github.io/swift-argument-parser/documentation/argumentparser/exitcode/issuccess

If so, I feel like it may be better to check `isSuccess` than test against `.failure`. Will this check assertion be satisfied for any nonzero failing exit code?

Tests/AirbnbSwiftFormatToolTests/AirbnbSwiftFormatToolTests.swift
```diff
@@ -74,26 +74,63 @@ final class AirbnbSwiftFormatToolTest: XCTestCase {
...
+        swiftLint: { _ in
+          ranSwiftLint = true
+          return EXIT_SUCCESS
+        },
+        swiftLintAutocorrect: { _ in
+          ranSwiftLintAutocorrect = true
+          return EXIT_SUCCESS
+        }))
+
+    XCTAssertEqual(error as? ExitCode, .failure)
```
bachand, line 133: This is a change in behavior, if I am understanding correctly.

bachand (commented):
@calda thanks for continuing to improve this tool! I was about to approve this PR but now I am realizing that I am unclear on what the behavior of our tool should be. Should the tool fail or succeed if there were violations that were automatically corrected? My understanding was that we wanted the tool to succeed if the violations were automatically corrected, which is the current behavior when running SwiftLint (see `testFormatWithOnlySwiftLintAutocorrectedViolation()`).

I am also recalling that we had a similar discussion previously in https://github.com/airbnb/swift/pull/242#discussion_r1355818517. In that discussion we aligned that we did not want the tool to fail with a nonzero exit code when errors could be automatically corrected. It seems like we are changing that behavior now, but only for SwiftFormat?

~In that previous discussion I suggested:~

> ~What do you think about documenting the exit code behavior of this tool in the README as part of this PR and then we can ensure that the behavior matches the documented behavior?~

~Personally, I think we should consider taking on that work as part of this PR. I am finding it challenging to review this PR without knowing what the expected behavior is. Moreover, I think this tool that we are building will be easier for others to adopt if we have documentation on how it is meant to function.~

-----

**Update:** We did update the README already with the expected behavior! 🎉 😍 I am sorry that I did not dig deeper and notice when I was writing this comment originally 🙇‍♂️ 


Screenshot 2024-04-17 at 2 51 21 PM

Based on this update documented behavior, though, it seems like we do not want to add the `--strict` flag 🤔 I am curious to get your thoughts @calda .

calda:
@bachand -- this is a change in behavior to match our internal `lint:swift` command`, which returns a non-zero exit code if SwiftFormat autocorrect applies any changes.

I was thinking it would make sense for this public tool to match the behavior of our internal tool, especially since we just added this `--strict` option to SwiftFormat for that use case.

You're right that we should also update the README, thanks for noticing. I'll make that change.

Do you think this change directionally makes sense, or do you think we should keep the current behavior where autocorrect mode will return a success / zero exit code when applying changes?

bachand:
> @bachand -- this is a change in behavior to match our internal `lint:swift` command, which returns a non-zero exit code if SwiftFormat autocorrect applies any changes.
> 
> I was thinking it would make sense for this public tool to match the behavior of our internal tool, especially since we just added this `--strict` option to SwiftFormat for that use case.
> 
> You're right that we should also update the README, thanks for noticing. I'll make that change.
> 
> Do you think this change directionally makes sense, or do you think we should keep the current behavior where autocorrect mode will return a success / zero exit code when applying changes?

@calda personally I think that documented behavior of this tool, returning 0 success code when any violations can be automatically corrected, makes sense. The motivation for returning a non-0 failure exit code in this case, as I understand, is so that we can fail Git pre-commit / pre-push hooks if there are automatically corrected violations that were made and now need to be committed. My personal belief is that by default this tool should return a successful code "when there is no action for the user to take", either because there were no violations or they have been automatically fixed. This aligns with the default behavior of SwiftFormat/SwiftLint, at least as of the last time that I [tested](https://github.com/airbnb/swift/pull/242#discussion_r1355831828) those tools.

That all being said, I think it would be reasonable to add a "strict" mode to our tool so that this tool can participate be used in a Git pre-commit / pre-push hook and fail that hook when autocorrections have been made. However, I think the strict mode should be opt-in.

I also think that it's important that the behavior of this tool is consistent. My understanding of this change is that the tool will fail if there's a SwiftFormat autocorrection but not if there is a SwiftLint autocorrection. I think that it's important that this tool has the same behavior for both "modes" of autocorrection.

calda:
Sounds good, thanks for sharing your thoughts. Let's leave this tool as-is for now then!

Package.swift
calda: I usually:
 1. Update the url to point to the new value
 2. Run `rm -rf .build && swift package format`
 3. SPM will print out an error that includes the old checksum and the new checksum
 4. Copy the new checksum into the package manifest

There's also a command to calculate this yourself but I don't know it off the top of my head so use this lazy method instead 😛

dfed: in case it's helpful: if you have the zip locally, you can always run `xcrun swift package compute-checksum <path_to_zip_file>`

PR #271, opened 2024-05-30
Remove "Platforms: macOS" supported platforms banner

calda:
We realized that the "Platforms: macOS" banner at the top of the README is potentially confusing.

These Swift Package Index banners refer to the Swift versions and platforms supported by our Swift package plugin. However, without this context it may look like this is saying that the Style Guide primarily supports macOS.

I still like the Swift versions banner, so think that's worth keeping. This has the side effect of suggesting that the style guide itself "supports" the given range of Swift versions, which is true.

### Before

image

### After

image

@airbnb/swift-styleguide-maintainers

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # Airbnb Swift Style Guide
 
-[![](https://img.shields.io/endpoint?url=https%3A%2F%2Fswiftpackageindex.com%2Fapi%2Fpackages%2Fairbnb%2Fswift%2Fbadge%3Ftype%3Dswift-versions)](https://swiftpackageindex.com/airbnb/swift)
-
 ## Goals
 
 Following this style guide should:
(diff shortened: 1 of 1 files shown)


bachand (approved):
I agree that we should remove this banner for clarity given its current placement.

I wonder if there is a way to clarify that the Swift version banner refers to the Swift package plugin?

Instead of removing this banner, we could move both banners from the top of the README to a section, or a standalone markdown file, about the package plugin 🤔

calda:
This is what it looks like if we move the badges to the package plugin subsection: 

image

I think this would be reasonable, but personally I prefer the look of having the supported version badge at the top of the page. After having the badge at the top for a while, the page without the badge at the top feels like it's missing something. I also like that it's a piece of dynamic content suggesting "yes, the style guide is 'up to date' for the latest version of Swift".

bachand:
Thanks for trying out the alternative idea, @calda . Keeping the Swift versions banner at the top and removing the platforms banner sounds good to me! I see what you are saying about it feeling like something is missing without the Swift versions banner at the top of the README.

PR #272, opened 2024-06-05
Add rule to remove blank lines between chained functions

mannylopez:
#### Summary

This PR proposes adding a rule that removes spaces between chained functions ([blankLinesBetweenChainedFunctions](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#blankLinesBetweenChainedFunctions) )

#### Reasoning

Improves readability and maintainability, making it easier to see the sequence of functions that are applied to the object.

#### Examples

```swift
// WRONG
var innerPlanetNames: [String] {
  planets
    .filter { $0.isInnerPlanet }

    .map { $0.name }
}

// RIGHT
 var innerPlanetNames: [String] {
  planets
    .filter { $0.isInnerPlanet }
    .map { $0.name }
}
  ```
**Note**

After https://github.com/nicklockwood/SwiftFormat/pull/1723 merges, the rule will also apply to cases where there is a comment between chained functions.

```swift
// WRONG
var innerPlanetNames: [String] {
  planets
    .filter { $0.isInnerPlanet }

    // Gets the name of the inner planet
    .map { $0.name }
}

// RIGHT
var innerPlanetNames: [String] {
  planets
    .filter { $0.isInnerPlanet }
    // Gets the name of the inner planet
    .map { $0.name }
}
  ```

#### Test

Confirmed that running the linter with `--rules blankLinesBetweenChainedFunctions` turned on does remove the extra space

image

### CODE REVIEW

File: Sources/AirbnbSwiftFormatTool/airbnb.swiftformat [hunk 1]
```diff
@@ -102,4 +102,5 @@
 --rules wrapMultilineConditionalAssignment
 --rules blankLineAfterMultilineSwitchCase
 --rules consistentSwitchStatementSpacing
---rules semicolons
\ No newline at end of file
+--rules semicolons
+--rules blankLinesBetweenChainedFunctions
```
mannylopez, line 106: Is this right?

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -102,4 +102,5 @@
 --rules wrapMultilineConditionalAssignment
 --rules blankLineAfterMultilineSwitchCase
```
mannylopez, line 103: `blankLineAfterMultilineSwitchCase` - Why can't I find this in either
- https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md or
- https://github.com/nicklockwood/SwiftFormat/blob/main/Sources/Rules.swift
??

.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
...
+        "swift"
+      ],
+      "args": [],
+      "cwd": "${workspaceFolder:swift}",
+      "name": "Release AirbnbSwiftFormatTool",
+      "program": "${workspaceFolder:swift}/.build/release/AirbnbSwiftFormatTool",
+      "preLaunchTask": "swift: Build Release AirbnbSwiftFormatTool"
+    }
+  ]
+}
```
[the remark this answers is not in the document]
mannylopez: I don't see `.vscode/launch.json` committed for a similar PR: https://github.com/airbnb/swift/pull/260/files which makes me think I should not include this.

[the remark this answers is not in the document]
calda: Sure, let's add `.vscode` to the `.gitignore`

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
calda: Some rules are still only available on the `develop` branch. You can see info about them in the `Rules.md` file on the `develop` branch: https://github.com/nicklockwood/SwiftFormat/blob/develop/Rules.md

SwiftFormat development happens on the `develop` branch. The `main` branch is only updated when releasing new versions.

README.md
```diff
@@ -2091,6 +2091,27 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='remove-blank-lines-between-chained-functions'></a>(<a href='#remove-blank-lines-between-chained-functions'>link</a>) **[Long](https://github.com/airbnb/swift#column-width) Remove blank lines between chained functions.**  [![SwiftFormat: blanklinesbetweenchainedfunctions](https://img.shields.io/badge/SwiftFormat-wrapArguments-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#blanklinesbetweenchainedfunctions)
+
+  <details>
+
+  ```swift
+  // WRONG
```
calda, line 2099: Perhaps we should also include a SwiftUI example, since SwiftUI is a very common use of method chaining (perhaps the most common!).

README.md
```diff
@@ -2091,6 +2091,27 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  [0, 1, 2]
```
calda, line 2100: Many examples in the style guide have an "outer space" theme to anchor the sample code with slightly less contrived examples. Perhaps we could come up with an example related to that theme? Something like `planets.filter { $0.isTerrestrialPlanet }.flatMap { $0.moons }` could be nice.

README.md
```diff
@@ -2091,6 +2091,27 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='remove-blank-lines-between-chained-functions'></a>(<a href='#remove-blank-lines-between-chained-functions'>link</a>) **Remove blank lines between chained functions.** [![SwiftFormat: blanklinesbetweenchainedfunctions](https://img.shields.io/badge/SwiftFormat-blankLinesBetweenChainedFunctions-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#blanklinesbetweenchainedfunctions)
```
mannylopez, line 2094: Preview of what the MarkDown looks like when it is rendered: https://gist.github.com/mannylopez/178ea1b36e97f55ede82cda2ed93b7b2

README.md
```diff
@@ -2091,6 +2091,27 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  <details>
+
+  ```swift
+  // WRONG
+  [0, 1, 2]
+    .map { $0 * 2 }
+
+
+
+    .map { $0 * 3 }
```
calda, line 2105: We already have the [consecutiveBlankLines](https://github.com/nicklockwood/SwiftFormat/blob/develop/Rules.md#consecutiveBlankLines) enabled -- let's limit the example to just a single blank line so it doesn't also conflict with that other rule.

```suggestion
    .map { $0 * 2 }

    .map { $0 * 3 }
```

README.md
calda, line 2094: Could we also include something related to the rationale? Many rules have a small discussion in a `#### Why?` section within the `<details>` block.

We would ideally include at least one sentence describing the rationale, both here in the README but also in the PR body.

calda: It's also easy to preview the markdown using this "Display the rich diff" button on the file in the diff view:

image

</br></br>

image

README.md
calda: It would also be nice to include something a property declaration `let innerMoons = ...`, since a bare value like this isn't valid code in most contexts.

README.md
mannylopez: TIL! Thanks, Cal!

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
mannylopez: thanks!

README.md
mannylopez: Updated!

README.md
```diff
@@ -2091,6 +2091,37 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  let planets = ["Mercury", "Venus", "Earth", "Mars"]
+
+  var transformedPlanetNames: [String] {
+      planets
+        .map { $0.uppercased() } // Making names uppercase
+        .map { $0 + "!" } // Adding excitement with an exclamation mark
+  }
+  ```
+
+  </details>
```
calda, line 2123: > The rule does not remove this empty space if there is a comment in between chained functions. Running the linter will leave this code as-is.
>
> ```swift
> let planets = ["Mercury", "Venus", "Earth", "Mars"]
> 
> var transformedPlanetNames: [String] {
>    planets
>      .map { $0.uppercased() } // Making names uppercase
>
>      // Adding excitement with an exclamation mark
>      .map { $0 + "!" }
> }
>  ```
  
Is this exception necessary? I'd prefer we not allow this, and for consistency remove the blank line here as well.

Looking at the [PR](https://github.com/nicklockwood/SwiftFormat/pull/1480) that introduced this SwiftFormat rule, I don't see any indication that this behavior was implemented intentionally. It would be easy to update the SwiftFormat rule to handle this case and remove this blank line as well.

What do you think @mannylopez?

README.md
```diff
@@ -2091,6 +2091,37 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  ```swift
+  // WRONG
+  let planets = ["Mercury", "Venus", "Earth", "Mars"]
+
+  var transformedPlanetNames: [String] {
+      planets
+        .map { $0.uppercased() } // Making names uppercase
+        
+        .map { $0 + "!" } // Adding excitement with an exclamation mark
+  }
```
calda, line 2111: Some thoughts on the example code:
 - Two sequential `map` calls is not common, since it would be simpler and more performant to just have a single map call like `map { $0.uppercased() + "!" }`. Instead, could we use `filter` and `map`?
 - I think we can omit the comments for simplicity (less is more!)
 - We can omit the `let planets = ["Mercury", "Venus", "Earth", "Mars"]` array since it isn't relevant to this specific rule.

```suggestion
  // WRONG
  var innerPlanetNames: [String] {
      planets
        .filter { $0.isInnerPlanet }
        
        .map { $0.name }
  }
```

We could _additionally_ have an example with code comments. That would help clarify the behavior of this [case](https://github.com/airbnb/swift/pull/272/files#r1631405303). I think it's good to address edge cases in separate examples, to keep the primary example nice and simple. For that example with comments, I think we should put the comments in between the method calls rather than on the right-hand side, since this formatting is more common imo.

README.md
```diff
@@ -2091,6 +2091,37 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  ```swift
+  // WRONG
+  let planets = ["Mercury", "Venus", "Earth", "Mars"]
+
+  var transformedPlanetNames: [String] {
+      planets
+        .map { $0.uppercased() } // Making names uppercase
+        
+        .map { $0 + "!" } // Adding excitement with an exclamation mark
```
calda, line 2110: Noticed there's too much indentation here

```suggestion
    planets
      .map { $0.uppercased() } // Making names uppercase
        
      .map { $0 + "!" } // Adding excitement with an exclamation mark
```

mannylopez: Good catch - I notice the 4 spaces, but assumed that it was the preference for this repo. Should have checked the other code examples, where it's clearly not the case. I'll update it.

README.md
mannylopez: I hear what you're saying and agree. The example you list is more realistic, so I'll include that.

README.md
mannylopez: Thanks for the help updating the rule. Here's the SwiftFormat PR to get rid of this exception: https://github.com/nicklockwood/SwiftFormat/pull/1723

mannylopez:
@calda https://github.com/nicklockwood/SwiftFormat/pull/1723 merged overnight. Do you think this PR is ready to merge as well?

calda:
We currently wait to merge until about one week after announcing the proposal review so everyone has a chance to share their feedback (so, we can merge this on Friday). We'll also want to update the SwiftFormat version in this repo to include your bug fix. I'm working on instructions for how to do that, which I'll share in a bit.

mannylopez:
> We currently wait to merge until about one week after announcing the proposal review so everyone has a chance to share their feedback (so, we can merge this on Thursday or Friday). We'll also want to update the SwiftFormat version in this repo to include your bug fix. I'm working on instructions for how to do that, which I'll share in a bit.

ah, that's right! Thanks.

calda (approved):
Thanks! Looks great, and very strong support on Slack.

PR #273, opened 2024-06-07
Add redundantProperty rule

calda:
#### Summary

This PR proposes a new rule to avoid defining properties that are then returned immediately. Instead, return the value directly.

Autocorrect support is implemented in the SwiftFormat `redundantProperty` rule, which as added in https://github.com/nicklockwood/SwiftFormat/pull/1654.

```swift
// WRONG
var spaceship: Spaceship {
  let spaceship = spaceshipBuilder.build(warpDrive: warpDriveBuilder.build())
  return spaceship
}

// RIGHT
var spaceship: Spaceship {
  spaceshipBuilder.build(warpDrive: warpDriveBuilder.build())
}

// WRONG
var spaceship: Spaceship {
  let warpDrive = warpDriveBuilder.build()
  let spaceship = spaceshipBuilder.build(warpDrive: warpDrive)
  return spaceship
}

// RIGHT
var spaceship: Spaceship {
  let warpDrive = warpDriveBuilder.build()
  return spaceshipBuilder.build(warpDrive: warpDrive)
}
```

#### Reasoning

Property declarations that are immediately returned are typically redundant and unnecessary. Sometimes these are unintentionally created as the byproduct of refactoring. Cleaning them up automatically simplifies the code. In some cases this also results in the `return` keyword itself being unnecessary, further simplifying the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3455,6 +3455,42 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='redundant-property'></a>(<a href='#redundant-property'>link</a>) **Avoid defining properties that are then returned immediately.** Instead, return the value directly. [![SwiftFormat: redundantProperty](https://img.shields.io/badge/SwiftFormat-redundantProperty-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantProperty)
+
+    <details>
+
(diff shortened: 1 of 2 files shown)


mannylopez (approved):
Makes sense to me. Can't think of a reason why we wouldn't want to simplify this.

PR #280, opened 2024-07-22
Add explicit category ordering configuration for organizeDeclarations rule

calda:
As of https://github.com/nicklockwood/SwiftFormat/pull/1736, the `organizeDeclarations` rule is now fully configurable, and lets you specify the sort order of declarations. The default behavior of the `organizeDeclarations` is now different from the behavior defined in our style guide. To keep the behavior we want, we can provide our own configuration explicitly.

The corresponding configuration that matches our style guide is:

```
--visibilityorder beforeMarks,instanceLifecycle,open,public,package,internal,fileprivate,private # organizeDeclarations
--typeorder nestedType,staticProperty,staticPropertyWithBody,classPropertyWithBody,instanceProperty,instancePropertyWithBody,staticMethod,classMethod,instanceMethod # organizeDeclarations
```

### DIFF
--- a/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
+++ b/Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
@@ -27,6 +27,7 @@
 --structthreshold 20 # organizeDeclarations
 --enumthreshold 20 # organizeDeclarations
 --organizetypes class,struct,enum,extension,actor # organizeDeclarations
+--typeorder nestedType,staticProperty,staticPropertyWithBody,classPropertyWithBody,instanceProperty,instancePropertyWithBody,staticMethod,classMethod,instanceMethod # organizeDeclarations
 --extensionacl on-declarations # extensionAccessControl
 --patternlet inline # hoistPatternLet
 --redundanttype inferred # redundantType, propertyType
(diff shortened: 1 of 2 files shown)


mannylopez (approved):
Thank you for keeping this up to date.

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -27,6 +27,8 @@
 --structthreshold 20 # organizeDeclarations
 --enumthreshold 20 # organizeDeclarations
 --organizetypes class,struct,enum,extension,actor # organizeDeclarations
+--visibilityorder beforeMarks,instanceLifecycle,open,public,package,internal,fileprivate,private # organizeDeclarations
+--typeorder nestedType,staticProperty,staticPropertyWithBody,classPropertyWithBody,instanceProperty,instancePropertyWithBody,staticMethod,classMethod,instanceMethod # organizeDeclarations
```
bachand, line 31: @calda is it expected that there is no `classProperty`?

bachand (commented):
Thank you @calda !

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
calda: There's no such thing as a stored class property (aka a class property without a body)! Found this out when I was first working on the `organizeDeclarations` rule: https://forums.swift.org/t/class-properties/16539

bachand: Ah! Good to know.

PR #283, opened 2024-08-01
Add emptyExtension rule

mannylopez:
#### Summary

This PR adds rule `emptyExtension` to remove empty extensions that define no properties, functions, or conformances.

 ```swift
// WRONG: The first extension is empty and redundant.
extension Planet {}

extension Planet: Equatable {}

// RIGHT: Empty extensions that add a protocol conformance aren't redundant.
extension Planet: Equatable {}
  ```

#### Reasoning

Improves readability since the code has no effect and should be removed for clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4003,7 +4003,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
   </details>
   
-* <a id='remove-empty-extensions'></a>(<a href='#remove-empty-extensions'>link</a>) **Remove empty, non-conforming, extensions.** Does not affect empty extensions that are appended by a macro. [![SwiftFormat: emptyExtension](https://img.shields.io/badge/SwiftFormat-emptyExtension-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#emptyExtension)
+* <a id='remove-empty-extensions'></a>(<a href='#remove-empty-extensions'>link</a>) **Remove empty extensions that define no properties, functions, or conformances.** [![SwiftFormat: emptyExtension](https://img.shields.io/badge/SwiftFormat-emptyExtension-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#emptyExtension)
 
   <details>
 
--- a/README.md
+++ b/README.md
@@ -4002,6 +4002,35 @@ _You can enable the following settings in Xcode by running [this script](resourc
   ```
 
   </details>
+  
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4002,6 +4002,35 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  extension String: Equatable {}
+  
+  @GenerateBoilerPlate
+  extension Foo {}
+
+  // RIGHT
+  extension String: Equatable {}
+  
+  @GenerateBoilerPlate
+  extension Foo {}
```
calda, line 4030: I think we can simplify the example in the style guide -- the examples here don't need to be exhaustive as the test cases in SwiftFormat. Maybe we also use a type that fits the space theme of other examples.

```suggestion
  // WRONG: The first extension is empty and redundant.
  extension Planet {}
  extension Planet: Equatable {}
  
  // RIGHT: Empty extensions that add a protocol conformance aren't redundant.
  extension Planet: Equatable {}
```

mannylopez: Updated!

README.md
```diff
@@ -4002,6 +4002,25 @@ _You can enable the following settings in Xcode by running [this script](resourc
   ```
 
   </details>
+  
+* <a id='remove-empty-extensions'></a>(<a href='#remove-empty-extensions'>link</a>) **Remove empty, non-conforming, extensions.** Does not affect empty extensions that are appended by a macro. [![SwiftFormat: emptyExtension](https://img.shields.io/badge/SwiftFormat-emptyExtension-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#emptyExtension)
```
mannylopez, line 4006: @calda 

Re: https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#emptyExtension

Should I like to the `develop` branch since this is not on `master` yet?

calda, line 4006: We could simplify this and improve clarity with something like:

```suggestion
* <a id='remove-empty-extensions'></a>(<a href='#remove-empty-extensions'>link</a>) **Remove empty extensions that define no properties, functions, or conformances.** [![SwiftFormat: emptyExtension](https://img.shields.io/badge/SwiftFormat-emptyExtension-008489.svg)](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#emptyExtension)
```

I don't think we need to mention macros here, since that is mostly hypothetical and is technically covered by "that define no properties or functions".

calda: So far we always just point to the `master` branch -- this link will be correct once the rule is officially released. Although, maybe we should just change all of our URLs to point to the `develop` branch. Either works for me!

mannylopez: Much better!

calda (approved):
Good internal support for this change, thanks @mannylopez!

PR #288, opened 2024-08-21
Add SwiftFormat void rule

miguel-jimenez-0529:
#### Summary

This PR replaces the SwiftLint `void_return` in favor to the SwiftFormat `void` rule in conjunction with the existing `redundantVoidReturnType` which cover the same functionality and additional rules described in the PR.

#### Reasoning

This is an effort to migrate all autocorrect rules to SwiftFormat

### DIFF
--- a/README.md
+++ b/README.md
@@ -3677,6 +3677,39 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='generic-void'></a>(<a href='#generic-void'>link</a>) **Avoid using `()` as a generic type**. Prefer `Void`.
+
+  <details>
+
+  ```swift
+  struct Wrapper<T> { ... }
+  // WRONG
+  let wrapper = Wrapper<()>()
+
+  // RIGHT
+  let wrapper = Wrapper<Void>()
+  ```
+  </details>
+
+* <a id='void-instance'></a>(<a href='#void-instance'>link</a>) **Avoid using `Void()` as an instance of `Void`**. Prefer `()`.
+
+  <details>
+
+  ```swift
+  struct Resolver<T> { 
+    func resolve(T) { ... }
+  }
+
+  let resolver = Resolver()
+
+  // WRONG
+  resolver.resolve(Void())
+
+  // RIGHT
+  resolver.resolve(())
+  ```
+  </details>
+
 **[⬆ back to top](#table-of-contents)**
 
 ## File Organization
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -3677,6 +3677,39 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='generic-void'></a>(<a href='#generic-void'>link</a>) **Avoid using `()` as a generic type**. Prefer `Void`.
```
calda, line 3680: Instead of saying "Avoid using `()` as a generic type" maybe we should say "Avoid using `()` as a _type_"

e.g. the rule would convert `let result: ()` to `let result: Void` (right?) but that's not covered by the "generic type" wording here

README.md
```diff
@@ -3677,6 +3677,39 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+
+  <details>
+
+  ```swift
+  struct Wrapper<T> { ... }
+  // WRONG
+  let wrapper = Wrapper<()>()
+
+  // RIGHT
+  let wrapper = Wrapper<Void>()
```
calda, line 3690: Maybe we use `Result` instead. `Result<Void, Error>` is pretty common.

```suggestion
  // WRONG
  let result: Result<(), Error>

  // RIGHT
  let result: Result<Void, Error>
```

README.md
```diff
@@ -3677,6 +3677,39 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    func resolve(T) { ... }
+  }
+
+  let resolver = Resolver()
+
+  // WRONG
+  resolver.resolve(Void())
+
+  // RIGHT
+  resolver.resolve(())
```
calda, line 3709: We could use `Result` here as well

```suggestion
  let completion: (Result<Void, Error>) -> Void 

  // WRONG
  completion(.success(Void()))

  // RIGHT
  completion(.success(()))
```

Sources/AirbnbSwiftFormatTool/swiftlint.yml
```diff
@@ -6,7 +6,6 @@ only_rules:
   - legacy_constructor
   - legacy_nsgeometry_functions
   - unused_optional_binding
-  - void_return # TODO: Replace with SwiftFormat void rule
```
calda, line 9: Thanks!

README.md
miguel-jimenez-0529: Good point, let me update it

calda (approved):
Awesome!

PR #289, opened 2024-08-22
Add SwiftUI properties sort

miguel-jimenez-0529:
#### Summary

This PR adds the SwiftUI property declaration sort rule to the style guide.

#### Reasoning

its preferred that SwiftUI properties are sorted by their dynamic property type. The order is determined by the first time a type appears:

```swift
    // WRONG
    struct CustomSlider: View {

      @Binding private var value: Value
      @State private var foo = Foo()
      @Environment(\.sliderStyle) private var style
      @State private var bar = Bar()
      @Environment(\.layoutDirection) private var layoutDirection

      private let range: ClosedRange<Double>
      private let step: Double.Stride
    }

    // RIGHT
    struct CustomSlider: View {

      @Binding private var value: Value
      @State private var foo = Foo()
      @State private var bar = Bar()
      @Environment(\.sliderStyle) private var style
      @Environment(\.layoutDirection) private var layoutDirection

      private let range: ClosedRange<Double>
      private let step: Double.Stride
    }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -4002,7 +4002,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
     }
     ```
 
-    Additioanlly, after grouping SwiftUI properties, its prefered that propeties are sorted by their dynamic property type. The order is determined by the first time a type appears:
+    Additioanlly, after grouping SwiftUI properties, its preferred that properties are sorted by their dynamic property type. The order is determined by the first time a type appears:
 
     ```swift
     // WRONG
--- a/README.md
+++ b/README.md
@@ -3908,7 +3908,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
   * Static properties
   * Static property with body
   * Class properties with body
-  * SwiftUI dynamic properties (@State, @Environment, @Binding, etc)
+  * SwiftUI dynamic properties (@State, @Environment, @Binding, etc), grouped by type
   * Instance properties
   * Instance properties with body
   * Static methods
@@ -4002,7 +4002,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
     }
     ```
 
-    Additioanlly, after grouping SwiftUI properties, its preferred that properties are sorted by their dynamic property type. The order is determined by the first time a type appears:
+    Additionally, within the grouping of SwiftUI properties, it is preferred that the properties are also grouped by their dynamic property type. The group order applied by the formatter is determined by the first time a type appears:
 
     ```swift
     // WRONG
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4001,6 +4001,37 @@ _You can enable the following settings in Xcode by running [this script](resourc
       private let step: Double.Stride
     }
     ```
+
+    Additioanlly, after grouping SwiftUI properties, its prefered that propeties are sorted by their dynamic property type. The order is determined by the first time a type appears:
```
calda, line 4005: The discussion about how the format rule behaves seems more like an implementation detail rather than a "style rule". We should probably either clarify that this is describing the behavior of the formatter:

```suggestion
    Additionally, after grouping SwiftUI properties, its preferred that properties are grouped by their dynamic property type. The sort order applied by the formatter is determined by the first time a type appears:
```

or we could 

```suggestion
    Additionally, after grouping SwiftUI properties, its preferred that properties are grouped by their dynamic property type.
```

Saying "grouped by their dynamic property type" rather than "sorted by" seems better here, because "grouped" is the user-facing style requirement and "sorting" is kinda just the implementation detail of how we enforce the grouping.

(There are also a few typos in this sentence, corrected them in the two suggestions)

calda, line 4005: In the description above the fold, should we change:

```
* SwiftUI dynamic properties (@State, @Environment, @Binding, etc)
```

to something like

```
* SwiftUI dynamic properties (@State, @Environment, @Binding, etc), grouped by property type
```

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -25,6 +25,7 @@
 --typeattributes prev-line # wrapAttributes
 --wrapternary before-operators # wrap
 --structthreshold 20 # organizeDeclarations
+--sortswiftuiprops first-appearance-sort
```
miguel-jimenez-0529, line 28: @calda I noticed there is a `#something` at the end of each, not sure what to add here.

calda: It's the name of the rule that uses the argument, so:

```suggestion
--sortswiftuiprops first-appearance-sort #organizeDeclarations
```

calda: I'd put the lower down below the `--typeorder` option

calda (approved):
Nice!

jqsilver:
Why not have a standardized order for property types?

miguel-jimenez-0529:
> Why not have a standardized order for property types?

@jqsilver we don't want to have a strong opinion on how folks define their modifiers but at the same time we want some consistency. This rule feels like a good balance between allowing you to define the order you want but ensuring a consistent order.

PR #292, opened 2024-09-30
Add rule to prefer a generated Equatable implementation when possible

calda:
#### Summary

This PR adds a new rule to prefer using a generated Equatable implementation when comparing all properties of a type. For structs, prefer using the compiler-synthesized Equatable implementation when possible.

This rule is implemented by the new `redundantEquatable` SwiftFormat rule added in https://github.com/nicklockwood/SwiftFormat/pull/1891.

```swift
/// WRONG: The `static func ==` implementation is redundant and error-prone.
struct Planet: Equatable {
  let mass: Double
  let orbit: OrbitalElements
  let rotation: Double

  static func ==(lhs: Planet, rhs: Planet) -> Bool {
    lhs.mass == rhs.mass
      && lhs.orbit == rhs.orbit
      && lhs.rotation == rhs.rotation
  }
}

/// RIGHT: The `static func ==` implementation is synthesized by the compiler.
struct Planet: Equatable {
  let mass: Double
  let orbit: OrbitalElements
  let rotation: Double
}

/// ALSO RIGHT: The `static func ==` implementation differs from the implementation that 
/// would be synthesized by the compiler and compared all properties, so is not redundant.
struct CelestialBody: Equatable {
  let id: UUID
  let orbit: OrbitalElements

  static func ==(lhs: Planet, rhs: Planet) -> Bool {
    lhs.id == rhs.id
  }
}
```

#### Reasoning

Manually-implemented Equatable implementations are verbose, and keeping them up-to-date is error-prone. For example, when adding a new property, it's possible to forget to update the Equatable implementation to compare it.


Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -67,6 +67,7 @@
 --rules redundantGet
 --rules redundantFileprivate
 --rules redundantRawValues
+--rules redundantEquatable
```
calda, line 70: Will update to a new SwiftFormat build that includes this rule before merging

miguel-jimenez-0529 (approved):

PR #293, opened 2024-10-22
Add rule to prefer a generated EnvironmentValue property implementation when possible

miguel-jimenez-0529:
#### Summary

This PR adds a new rule to prefer using a generated EnvironmentValues property implementation.

This rule is implemented by the new `environmentEntry` SwiftFormat rule added in https://github.com/nicklockwood/SwiftFormat/pull/1908.

 ```swift
 /// WRONG: The `EnvironmentValues` property depends on `IsSelectedEnvironmentKey`
 struct IsSelectedEnvironmentKey: EnvironmentKey {
   static var defaultValue: Bool { false }
 }
 
 extension EnvironmentValues {
   var isSelected: Bool {
    get { self[IsSelectedEnvironmentKey.self] }
    set { self[IsSelectedEnvironmentKey.self] = newValue }
   }
 }

/// RIGHT: The `EnvironmentValues` property uses the @Entry macro 
extension EnvironmentValues {
  @Entry var isSelected: Bool = false
}
```

#### Reasoning

 Manually-implemented environment keys are verbose and it is considered a legacy pattern. `@Entry` was specifically intended to be a replacement considering it was back ported to iOS 13.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3745,6 +3745,35 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='redundant-environment-key-implementation'></a>(<a href='#redundant-environment-key-implementation'>link</a>) **Prefer using the `@Entry` macro to define properties inside `EnvironmentValues`**. When adding properties to SwiftUI `EnvironemtnValues`, prefer using the compiler-synthesized property implementation when possible. [![SwiftFormat: environmentEntry](https://img.shields.io/badge/SwiftFormat-environmentEntry-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/develop/Rules.md#environmentEntry)
+
+    <details>
+
+    ### Why?
+
+    Manually-implemented environment keys are verbose, and keeping them up-to-date is error-prone.
+
+    ```swift
+    /// WRONG: The `EnvironmentValues` property depends on `IsSelectedEnvironmentKey`
+    struct IsSelectedEnvironmentKey: EnvironmentKey {
+      static var defaultValue: Bool { false }
+    }
+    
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -3745,6 +3745,35 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
     </details>
 
+* <a id='redundant-environment-key-implementation'></a>(<a href='#redundant-environment-key-implementation'>link</a>) **Prefer using the `@Entry` macro to define properties inside `EnvironmentValues`**. When adding properties to SwiftUI `EnvironemtnValues`, prefer using the compiler-synthesized property implementation when possible. [![SwiftFormat: environmentEntry](https://img.shields.io/badge/SwiftFormat-environmentEntry-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/develop/Rules.md#environmentEntry)
+
+    <details>
+
+    ### Why?
+
+    Manually-implemented environment keys are verbose, and keeping them up-to-date is error-prone.
```
calda, line 3754: Just curious, which part of this do you think is error-prone about keeping these up-to-date? I do definitely agree that the old pattern is much more verbose.

Another thing to mention could be that manually implementing `EnvironmentKey` is a legacy pattern and `@Entry` was specifically intended to be a replacement.

calda (commented):
Remember to update `airbnb.swiftformat` and the SwiftFormat build before we merge this!

README.md
miguel-jimenez-0529: You are right, it may not be super error-prone. Let me update the details

calda (approved):
Love it! Nice work!

PR #294, opened 2024-12-09
Update to SwiftFormat 0.56-beta-1 based on 0.55 release

calda:
SwiftFormat 0.55 was recently released. This PR updates our build to 0.56-beta-1, which is based off of 0.55.3.

Several rules that were previously only available on `develop` were released officially in 0.55.0. As a part of this some of the rule and option names were updated.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,23 +7,7 @@ on:
     branches: [ master ]
 
 jobs:
-  test-package-plugin-macos-12:
-    name: Test Package Plugin
-    runs-on: macos-12
-    strategy:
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -7,23 +7,7 @@ on:
     branches: [ master ]
 
 jobs:
-  test-package-plugin-macos-12:
```
calda, line 10: The macOS 12 runner is now deprecated so needs to be removed:

image

miguel-jimenez-0529 (approved):
LGTM

PR #304, opened 2025-05-01
Remove unusedPrivateDeclarations rule

mannylopez:
#### Summary

We are removing the `unusedPrivateDeclarations` rule from the Airbnb Swift Style Guide

#### Reasoning

Limits the capabilities of Swift macros and creates more edge cases since linting is destructive.


calda (approved):
Thanks!

PR #305, opened 2025-05-02
Update SwiftLint to 0.59.1

makwanbarzan:
#### Summary

This PR updates SwiftLint to the latest version, as of now, which is 0.59.1.


calda:
Thanks for the PR -- this repo tracks the tool versions that we use within Airbnb, and we still use 0.55.1 internally. I looked at updating us to use 0.59.1, but there are some breaking changes with our setup that would need to be resolved.

PR #309, opened 2025-05-16
Cal  wrap string interpolation

calda:
By default SwiftFormat will insert linebreaks in multi-line string interpolations. This is allowed by the language and doesn't affect the resulting string values, but in practice produces ugly results that hurts readability.

This PR enables `--wrapstringinterpolation preserve` option, which was added in https://github.com/nicklockwood/SwiftFormat/pull/2059.

Since we don't have a style rule specifically discussing line wrapping in string interpolation, and we just got this behavior by default when enabling the SwiftFormat wrapping behavior, I don't feel the need to address this case directly in the style guide.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -42,7 +42,8 @@ namespace :update do
     .binaryTarget(
       name: "swiftformat",
       url: "https://github.com/calda/SwiftFormat-nightly/releases/download/#{latest_version_number}/SwiftFormat.artifactbundle.zip",
-      checksum: "#{checksum.strip}"),
+      checksum: "#{checksum.strip}"
+    ),
     EOS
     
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -42,7 +42,8 @@ namespace :update do
     .binaryTarget(
       name: "swiftformat",
       url: "https://github.com/calda/SwiftFormat-nightly/releases/download/#{latest_version_number}/SwiftFormat.artifactbundle.zip",
-      checksum: "#{checksum.strip}"),
+      checksum: "#{checksum.strip}"
+    ),
```
calda, line 46: While we're here, update the `update:swiftformat` rake task to output correctly-formatted code

PR #313, opened 2025-06-11
add new proposed rule in new Testing section

jqsilver:
#### Summary

This PR proposes a new rule in our style guide: Use `XCTUnwrap` or `#require` with optionals instead of `guard let` or any other unwrapping patterns. 

#### Reasoning

These APIs are much simpler and natural for unwrapping an optional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4369,7 +4369,7 @@ _You can enable the following settings in Xcode by running [this script](resourc
     @Test
     func something() throws {
       // RIGHT:
-      let value = #require(optinalValue)
+      let value = try #require(optionalValue)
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -4340,6 +4340,49 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  ```
+
+  ```swift
+  import Testing
+
+  struct SomeTests {
+    @Test
+    func something() throws {
+      // RIGHT:
+      let value = #require(optinalValue)
```
```suggestion
      let value = try #require(optionalValue)
```

PR #314, opened 2025-06-11
add new rule to prefer throwing in tests over `try!`

jqsilver:
#### Summary

Avoid forced `try!` and instead make test methods `throws`. 

#### Reasoning

XCTest and Swift Testing handle thrown errors if the test method is marked as throwing. `try!` is especially bad since it crashes the test, and we have a new formatting rule for it.


calda (approved):
awesome!

PR #317, opened 2025-06-25
Add urlMacro rule

mannylopez:
#### Summary

This PR adds rule `urlMacro` to prefer using `#URL(_:)` macro, if available in your project, instead of force-unwrapping `URL(string:)!` initializer

With `--urlmacro #URL,URLFoundation`:
```diff
- let url = URL(string: "https://example.com")!
+ import URLFoundation
+ let url = #URL("https://example.com")
```

#### Reasoning

Removes force-unwrap and potential for a crash


Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -47,6 +47,7 @@
 --shortoptionals always # typeSugar
 --semicolons never # semicolons
 --doccomments preserve # docComments
+--urlmacro #URL,URLFoundation
```
calda: Since `URLFoundation` is only available internally, we shouldn't actually enable this option in the `airbnb.swiftformat` here in this repo

```suggestion
```

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
```diff
@@ -128,3 +129,4 @@
 --rules modifiersOnSameLine
 --rules throwingTests
 --rules noGuardInTests
+--rules urlMacro
```
calda:
```suggestion

```

README.md
```diff
@@ -3882,6 +3882,24 @@ _You can enable the following settings in Xcode by running [this script](resourc
   ```
   </details>
 
+* <a id='url-macro'></a>(<a href='#url-macro'>link</a>) **Prefer the configured `#URL(_:)` macro instead of force-unwrapping `URL(string:)!` initializer`**. [![SwiftFormat: urlMacro](https://img.shields.io/badge/SwiftFormat-urlMacro-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/develop/Rules.md#urlMacro)
```
calda: "Configured" isn't necessary here.

Although since this relies on an internal tool, we should specify something like "If available..."

```suggestion
* <a id='url-macro'></a>(<a href='#url-macro'>link</a>) If available in your project, **prefer using a `#URL(_:)` macro instead of force-unwrapping `URL(string:)!` initializer`**. [![SwiftFormat: urlMacro](https://img.shields.io/badge/SwiftFormat-urlMacro-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/develop/Rules.md#urlMacro)
```

Sources/AirbnbSwiftFormatTool/airbnb.swiftformat
mannylopez: Ah, gotcha - that makes complete sense. Removing now.

README.md
```diff
@@ -3882,6 +3882,24 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+    #### Why?
+
+    The `#URL` macro provides compile-time validation of URL strings, eliminating runtime crashes from invalid URLs while maintaining clean syntax for static URL creation.
+
+    ```swift
+    // WRONG
+    let url = URL(string: "https://example.com")!
+
+    // RIGHT
+    import URLFoundation
```
calda: We can leave out this part since it's only relevant internally

```suggestion
```

PR #318, opened 2025-06-25
Add RedundantMemberwiseInit rule

miguel-jimenez-0529:
#### Summary

This PR adds rule `RedundantMemberwiseInit` to automatically remove redundant memberwise initializers from structs when they only assign parameters to properties with matching names.

With `--redundantmemberwise`:

```swift
- struct Planet {
-   let name: String
-   let mass: Double
-   
-   init(name: String, mass: Double) {
-     self.name = name
-     self.mass = mass
-   }
- }

+ struct Planet {
+   let name: String
+   let mass: Double
+ }
```

#### Reasoning

Removing redundant memberwise initializers reduces boilerplate code and improves readability. The compiler can synthesize equivalent memberwise initializers automatically for structs, so explicit initializers that only perform simple property assignment are unnecessary. This rule helps maintain cleaner, more concise code while preserving the same functionality.

🤖 Generated with [Claude Code](https://claude.ai/code)


README.md
```diff
@@ -2709,6 +2709,50 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 * <a id='time-intensive-init'></a>(<a href='#time-intensive-init'>link</a>) **Avoid performing any meaningful or time-intensive work in `init()`.** Avoid doing things like opening database connections, making network requests, reading large amounts of data from disk, etc. Create something like a `start()` method if these things need to be done before an object is ready for use.
 
+* <a id='omit-redundant-memberwise-init'></a>(<a href='#omit-redundant-memberwise-init'>link</a>) **Omit redundant memberwise initializers.** The compiler can synthesize memberwise initializers for structs, so explicit initializers that only assign parameters to properties with the same names should be omitted. [![SwiftFormat: redundantMemberwiseInit](https://img.shields.io/badge/SwiftFormat-redundantMemberwiseInit-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#redundantMemberwiseInit)
```
mannylopez: The rule does not exist on the `main` branch of SwiftFormat, but does exist in the `develop` branch, so this URL can be updated to

https://github.com/nicklockwood/SwiftFormat/blob/develop/Rules.md#redundantMemberwiseInit

miguel-jimenez-0529: Good point, @calda should we point to `develop` or `main` is acceptable considering there will be a new release in the SwiftFormat repo eventually?

calda: We've always just pointed urls to the `main` branch, once the rule is officially released the URL will work correctly

mannylopez: I'll update my PR then

calda: Worth mentioning that this only applies to internal inits

```suggestion
* <a id='omit-redundant-memberwise-init'></a>(<a href='#omit-redundant-memberwise-init'>link</a>) **Omit redundant memberwise initializers.** The compiler synthesize internal memberwise initializers for structs, so explicit internal initializers that only assign parameters to properties with the same names should be omitted. [![SwiftFormat: redundantMemberwiseInit](https://img.shields.io/badge/SwiftFormat-redundantMemberwiseInit-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#redundantMemberwiseInit)
```

README.md
```diff
@@ -2709,6 +2709,50 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  }
+
+  // RIGHT
+  struct Planet {
+    let name: String
+    let mass: Double
+    let radius: Double
+  }
+
+  // ALSO RIGHT: Custom logic in initializer makes it non-redundant
```
calda, line 2740: Maybe we should include an `ALSO RIGHT:` case with a public struct and a public init

README.md
```diff
@@ -2709,6 +2709,50 @@ _You can enable the following settings in Xcode by running [this script](resourc
...
+  struct Planet {
+    let name: String
+    let mass: Double
+    let radius: Double
+
+    init(name: String, mass: Double, radius: Double) {
+      self.name = name.capitalized
+      self.mass = max(0, mass)
+      self.radius = max(0, radius)
```
calda, line 2749: nice example 👌🏻

calda:
Remember to update `airbnb.swiftformat`

README.md
miguel-jimenez-0529: I think it can apply to `private` and `fileprivate` too. Let me update

PR #376, opened 2026-03-26
Add cifilter-builtins style guide rule

stephenfurlani-cell:
Prefer CIFilter typed factory methods (CIFilterBuiltins) over string-based CIFilter(name:) and setValue(_:forKey:). Adds rule entry to the Patterns section immediately before File Organization.

#### Summary

Adds a new **Patterns** rule recommending `CIFilter`'s typed factory methods (from `CIFilterBuiltins`) over the string-based `CIFilter(name:)` initializer and KVO `setValue(_:forKey:)`.

No SwiftLint or SwiftFormat rule exists for this — it is a style guide entry only.

#### Reasoning

`CIFilter(name:)` returns an optional and requires stringly-typed KVO for parameter access. The `CIFilterBuiltins` factory methods (available since iOS 14) provide non-optional returns and compile-time type-checked properties, eliminating an entire class of runtime errors.

The typed API requires `import CoreImage.CIFilterBuiltins`, which is an unusual submodule import but is Apple's documented approach.

Reference: https://developer.apple.com/documentation/coreimage/processing-an-image-using-built-in-filters

### DIFF
--- a/README.md
+++ b/README.md
@@ -3884,6 +3884,50 @@ _You can enable the following settings in Xcode by running [this script](resourc
 
 **[⬆ back to top](#table-of-contents)**
 
+* <a id='cifilter-builtins'></a>(<a href='#cifilter-builtins'>link</a>) **Prefer CIFilter's typed factory methods (via `CIFilterBuiltins`) over the string-based `CIFilter(name:)` initializer and KVO `setValue(_:forKey:)`.**
+
+  <details>
+
+  #### Why?
+
+  The typed factory methods introduced in iOS 14 (`import CoreImage.CIFilterBuiltins`) return non-optional, concrete filter objects with strongly-typed properties. This eliminates:
+
+  - The failable `CIFilter(name:)` initializer, which returns `nil` for typos and requires a `guard`/`if-let`.
+  - `setValue(_:forKey:)` calls that accept `Any?` and crash at runtime on wrong types or misspelled keys.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4381,27 +3880,48 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+  kMeansFilter.setValue(ciImage, forKey: kCIInputImageKey)
+  kMeansFilter.setValue(CIVector(cgRect: ciImage.extent), forKey: "inputExtent")
+  kMeansFilter.setValue(1, forKey: "inputCount")
+  kMeansFilter.setValue(5, forKey: "inputPasses")
+  ```
+
+  ```swift
   // RIGHT
-  let url = #URL("https://example.com")
+  import CoreImage.CIFilterBuiltins
```
calda: Can this be just `import CoreImage` or do you _need_ to specifically import `CoreImage.CIFilterBuiltins` in addition to `CoreImage`?

calda (approved):
Looks nice thanks! Could you please move it to the new "Apple Frameworks" section of the style guide, and run the markdown linter? (`bundle exec rake lint:markdown`)

README.md
calda, line 4429: Put the rule above the section's `back to top` button (as opposed to adding a new one)

README.md
stephenfurlani-cell: you need to do the sub-module, as it's not included in the general CoreImage framework header.

PR #378, opened 2026-04-16
Add "Apple Frameworks" section

calda:
This PR adds a new "Apple Frameworks" section. This gives us a place to put rules like https://github.com/airbnb/swift/pull/376.

### DIFF
--- a/site/site_content.rb
+++ b/site/site_content.rb
@@ -152,6 +152,9 @@ class SiteContent
       stripped.start_with?('[![](') && stripped.include?('swiftpackageindex.com')
     end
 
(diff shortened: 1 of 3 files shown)


site/site_content.rb
```diff
@@ -152,6 +152,9 @@ def filter_readme(
       stripped.start_with?('[![](') && stripped.include?('swiftpackageindex.com')
     end
 
+    # Exclude the swift.airbnb.tech URL line from the site.
+    lines = lines.reject { |line| line.strip == 'https://swift.airbnb.tech' }
```
calda, line 156: Unrelated, meant to filter out this line in the readme from the site content:

Image

PR #382, opened 2026-06-05
Add rule to avoid calling `State(wrappedValue:)` initializer directly

calda:
#### Summary

This PR adds a new rule to avoid calling the `State(wrappedValue:)` initializer explicitly, in favor of just setting the wrapped property directly.

#### Reasoning

Calling `State(wrappedValue:)` or `State(initialValue)` explicitly is always unnecessary. The [SwiftUI docs](https://developer.apple.com/documentation/swiftui/state/init(wrappedvalue:)) say: "You don’t call this initializer directly".

#### Examples

  ```swift
  // WRONG
  struct SpacecraftView: View {
    init(telemetryService: TelemetryService) {
      _spacecraftName = .init(wrappedValue: "Voyager")
      _telemetryService = .init(wrappedValue: telemetryService)
    }

    @State private var spacecraftName: String
    @ObservedObject private var telemetryService: TelemetryService
  }

  // RIGHT
  struct SpacecraftView: View {
    init(telemetryService: TelemetryService) {
      spacecraftName = "Voyager"
      self.telemetryService = telemetryService
    }

    @State private var spacecraftName: String
    @ObservedObject private var telemetryService: TelemetryService
  }
  ```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -47,8 +47,8 @@ let package = Package(
 
     .binaryTarget(
       name: "swiftformat",
-      url: "https://github.com/calda/SwiftFormat-nightly/releases/download/2026-06-10/SwiftFormat.artifactbundle.zip",
-      checksum: "3998613fa1b26c1f70d9d13260eff43fbaeb916a8fd2183419814da5efbcd8b7"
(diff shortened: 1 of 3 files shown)


calda:
We found cases internally where this actually does cause a runtime behavior change. This stuff is complicated!

PR #385, opened 2026-06-26
Add rule: prefer contains over range(of:) compared against nil

jparise:
#### Summary

Enables SwiftFormat's `preferContainsOverRange` rule and documents the corresponding style-guide entry (sibling to the existing `count(where:)`, `isEmpty`, and `flatMap` collection rules). Also bumps the SwiftFormat artifactbundle to the `2026-06-26` nightly, the first release to include the rule.

#### Reasoning

`range(of:) != nil` computes a range only to discard it and test for membership; `contains` expresses that membership check directly and reads more clearly (`range(of:) == nil` becomes `!contains`).

The rule is conservative — only a single `of:` argument is matched, and the `== nil` negation bails on optional-chained receivers and other cases where a prefix `!` can't be placed safely. It ships disabled-by-default (in rare cases the rewrite can change behavior, e.g. an `NSString` receiver whose `range(of:)` returns a non-optional `NSRange` where `!= nil` is always true); listing it in `--rules` enables it, as we already do for `isEmpty`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -47,8 +47,8 @@ let package = Package(
 
     .binaryTarget(
       name: "swiftformat",
-      url: "https://github.com/calda/SwiftFormat-nightly/releases/download/2026-06-24/SwiftFormat.artifactbundle.zip",
-      checksum: "77cd21d4d4218ba741745848e60c184ac72357717c0ab67722c33c4cdb654ea6"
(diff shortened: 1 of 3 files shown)


calda (approved):
Thanks!

calda:
@copilot please resolve merge conflicts

PR #386, opened 2026-06-26
Add rule: prefer first(where:) over filter(_:).first

jparise:
#### Summary

Enables SwiftFormat's `preferFirstWhere` rule and documents the corresponding style-guide entry (sibling to the existing `count(where:)`, `isEmpty`, and `flatMap` collection rules). Also bumps the SwiftFormat artifactbundle to the `2026-06-26` nightly, the first release to include the rule.

#### Reasoning

`filter { ... }.first` builds an entire filtered collection just to take its first element; `first(where:)` finds that element directly and stops at the first match. The rule only matches the no-argument `.first` property (leaving `.first(where:)` / `.first(3)` method calls alone), so it's a behavior-preserving rewrite.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -47,8 +47,8 @@ let package = Package(
 
     .binaryTarget(
       name: "swiftformat",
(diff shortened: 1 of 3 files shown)


calda (approved):
Thanks!

PR #387, opened 2026-06-30
Add rule: prefer min() over sorted().first

jparise:
#### Summary

Enables SwiftFormat's `preferMinOverSorted` rule and documents the corresponding style-guide entry (sibling to the existing `count(where:)`, `isEmpty`, `flatMap`, `preferContainsOverRange`, and `preferFirstWhere` collection rules). Also bumps the SwiftFormat artifactbundle to the `2026-06-30` nightly, the first release to include the rule.

#### Reasoning

`sorted().first` sorts the entire sequence — `O(n log n)` plus a full copy — just to take the smallest element; `min()` finds it in a single `O(n)` pass with no allocation. `sorted(by:).first` likewise becomes `min(by:)`.

Only `.first` is rewritten. `sorted().last` → `max()` is *not* behavior-preserving on comparator ties: both `min`/`max` keep the *first* element of a tie group, but `sorted().last` reads the *last*, so any element with identity beyond the comparator key (tuples-by-key, objects) would silently change. The rule intentionally leaves `.last` alone, so it's a behavior-preserving rewrite.

Enabling the rule also autocorrected one site in this repo (`Plugin.swift`: `supportedSwiftVersions.sorted().first!` → `.min()!`).

### DIFF
--- a/Plugins/FormatSwift/Plugin.swift
+++ b/Plugins/FormatSwift/Plugin.swift
@@ -196,7 +196,7 @@ extension Package {
 
   /// The minimum Swift version supported by this package
   var minimumSwiftVersion: SwiftVersion {
-    supportedSwiftVersions.sorted().first!
+    supportedSwiftVersions.min()!
   }
 
   // MARK: Private
--- a/README.md
+++ b/README.md
@@ -4488,6 +4488,28 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   </details>
 
+- <a id='prefer-min-over-sorted'></a>(<a href='#prefer-min-over-sorted'>link</a>) **Prefer using `min()` over `sorted().first`**.
+
+  <details>
+
+  [![SwiftFormat: preferMinOverSorted](https://img.shields.io/badge/SwiftFormat-preferMinOverSorted-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#preferMinOverSorted)
+
+  #### Why?
+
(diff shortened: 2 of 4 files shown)


Plugins/FormatSwift/Plugin.swift
```diff
@@ -196,7 +196,7 @@ extension Package {
 
   /// The minimum Swift version supported by this package
   var minimumSwiftVersion: SwiftVersion {
-    supportedSwiftVersions.sorted().first!
```
calda, line 199: hah! nice fix

README.md
```diff
@@ -4488,6 +4488,30 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+  <details>
+
+  [![SwiftFormat: preferMinOverSorted](https://img.shields.io/badge/SwiftFormat-preferMinOverSorted-7B0051.svg)](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#preferMinOverSorted)
+
+  #### Why?
+
+  `sorted().first` sorts the entire sequence — `O(n log n)` plus a full copy — just to take the smallest element. `min()` finds it in a single `O(n)` pass with no allocation. `sorted(by:).first` likewise becomes `min(by:)`.
+
+  Only `.first` is rewritten: `sorted().last` → `max()` is _not_ behavior-preserving on comparator ties (`min`/`max` keep the first tied element, but `sorted().last` reads the last), so it is intentionally left alone.
+
```
calda: This aside feels unrelate to the prose rule itself

```suggestion
```

calda (approved):

README.md
jparise: Done — removed the aside in 5ffec67. (The `.last` → `max()` tie-divergence rationale lives in the upstream commit/PR if anyone needs it.)

PR #388, opened 2026-06-30
Add rule: prefer contains(where:) over first(where:) compared against nil

jparise:
#### Summary

Enables SwiftFormat's `preferContainsOverFirst` rule and documents the corresponding style-guide entry (sibling to the existing `count(where:)`, `isEmpty`, `flatMap`, `preferContainsOverRange`, and `preferFirstWhere` collection rules). Also bumps the SwiftFormat artifactbundle to the `2026-06-30` nightly, the first release to include the rule.

#### Reasoning

`first(where:) != nil` (and `firstIndex(where:) != nil`) finds and returns the matched element or index only to discard it and test for existence; `contains(where:)` expresses that existence check directly and returns a `Bool`. The `== nil` forms become the negated `!contains(where:)`. `firstIndex(where:)` is equivalent because an index exists exactly when a matching element exists.

The rule is conservative — it matches only `where:`-closure calls (leaving `firstIndex(of:)`, the bare `.first` property, and `first(n)` alone), only `!= nil` / `== nil` comparisons, and the `== nil` negation bails on optional-chained receivers and other cases where a prefix `!` can't be placed safely. So it's a behavior-preserving rewrite.

Pairs with the existing `preferContainsOverRange` rule (same `!= nil` / `== nil` → `contains` / `!contains` shape).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -47,8 +47,8 @@ let package = Package(
 
     .binaryTarget(
       name: "swiftformat",
-      url: "https://github.com/calda/SwiftFormat-nightly/releases/download/2026-06-26/SwiftFormat.artifactbundle.zip",
-      checksum: "94fd5f5bf9d17dfbe4a687cda40302a2e3a70e853a3dc27c9ac58d64f3ab175b"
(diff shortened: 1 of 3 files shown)


calda (approved):
Nice!

calda:
@copilot please resolve merge conflicts

PR #393, opened 2026-07-09
Add rule to reuse the existing identifier when unwrapping an optional

calda:
#### Summary

This PR extends the optional binding rule to cases where the lhs and rhs identifiers are different: "when unwrapping an optional, prefer reusing the existing identifier rather than introducing a new identifier."

```swift
// NOT PREFERRED
if let ship = spaceship {
  launch(ship)
}

// PREFERRED
if let spaceship {
  launch(spaceship)
}
```

#### Reasoning

Reusing the optional's existing name, rather than introducing a new identifier for the unwrapped value, is more idiomatic and reduces cognitive load by avoiding introducing new variable names for the same object.

However, since there is nuance, we will not enforce this with code autocorrect.

>  It's fine to introduce a new name, however, when doing so improves clarity at the use site:
>
>  ```swift
>  // ALSO RIGHT
>  if let destination = pendingDestination {
>    spaceship.travel(to: destination)
>  }
>  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1136,37 +1136,37 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   <details>
 
-  [![SwiftFormat: redundantOptionalBinding](https://img.shields.io/badge/SwiftFormat-redundantOptionalBinding-7B0051.svg)](http://swiftformat.info/rules/prerelease#redundantOptionalBinding)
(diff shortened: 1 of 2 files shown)


calda:
@copilot please add `--redundant-optional-binding always #redundantOptionalBinding` to `airbnb.swiftformat` and also fix the Lint Markdown failure.

PR #399, opened 2026-07-31
Add rule to prefer if expressions over ternary operators in single-expression return values

calda:
#### Summary

**Prefer `if` expressions over ternary operators for single-expression return values.** Use ternary operators for conditions nested in other expressions, such as SwiftUI modifier conditions. Generally prefer `if` expressions for assignments after `=` operators.

  ```swift
  // WRONG
  var destination: Planet {
    spaceship.hasWarpDrive
      ? .proximaCentauri
      : .mars
  }

  // RIGHT
  var destination: Planet {
    if spaceship.hasWarpDrive {
      .proximaCentauri
    } else {
      .mars
    }
  }

  // ALSO RIGHT. Use ternaries for conditions nested in other expressions,
  // like SwiftUI modifier conditions.
  Image(.spaceship)
    .foregroundStyle(spaceship.velocity > 0 ? .orange : .blue)
  ```

#### Reasoning

If expressions are easier to read than ternary operators, especially when nesting multiple conditions. If expressions are more modern and support more functionality, like `if let` conditions.


README.md
```diff
@@ -1476,6 +1476,47 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+  // RIGHT
+  var destination: Planet {
+    if spaceship.hasWarpDrive {
+      .proximaCentauri
+    } else {
+      .mars
+    }
+  }
+
+  // ALSO RIGHT. Use ternaries for conditions nested in other expressions,
```
calda, line 1513: @copilot, please add an "ALSO RIGHT" for a computed var ternary where the ternary is on a single line rather than wrapped. Single-line ternaries are permitted. Use the color example.

README.md
```diff
@@ -1476,6 +1476,47 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
 
+- <a id='prefer-if-expressions-over-ternary-operators'></a>(<a href='#prefer-if-expressions-over-ternary-operators'>link</a>) **Prefer `if` expressions over ternary operators for single-expression return values.** Use ternary operators for conditions nested in other expressions, such as SwiftUI modifier conditions. Generally prefer `if` expressions for assignments after `=` operators.
+
+  <details>
+
+  <!-- ai-skill-include: autocorrect doesn't apply to conditional assignments after `=`, but this is still a good best practice -->
+
+  [![SwiftFormat: ifExpressions](https://img.shields.io/badge/SwiftFormat-ifExpressions-7B0051.svg)](https://swiftformat.info/rules/prerelease#ifExpressions)
+
+  ```swift
```
calda, line 1491: @copilot add a `### Why` like other rules with: If expressions are more readable than ternary expressions, especially for multiple nested conditions or multi-line values.

README.md
```diff
@@ -1476,6 +1476,50 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+  // RIGHT
+  var destination: Planet {
+    if spaceship.hasWarpDrive {
+      .proximaCentauri
+    } else {
+      .mars
+    }
+  }
+
+  // ALSO RIGHT. Single-line ternaries are permitted.
```
calda, line 1508: @copilot, wrap after the `{`, but the ternary itself can be on a single line.

PR #402, opened 2026-08-20
Add rule: prefer lazy.map before single-pass operations

jparise:
#### Summary

Enables SwiftFormat's `preferLazyMap` rule and documents the corresponding style-guide entry (a sibling of the existing `preferMinOverSorted`, `preferFirstWhere`, and `preferContains` collection rules).

#### Reasoning

`map` allocates an array of every transformed element. When all that happens to that array is a single walk over it, that allocation is waste — `lazy.map` transforms each element as it's visited instead.

```swift
// WRONG
let names = users.map { $0.name }.joined(separator: ", ")
let minY = vertices.map { $0.y }.min()

// RIGHT
let names = users.lazy.map { $0.name }.joined(separator: ", ")
let minY = vertices.lazy.map { $0.y }.min()
```

Qualifying operations consume their receiver exactly once (`min`, `max`, `reduce`, `joined(separator:)`) or can additionally stop early and skip transforming the rest (`contains`, `allSatisfy`, `first(where:)`). The style-guide entry notes the converse too: `lazy` is only right when the result really is consumed once, since `sorted()` materializes anyway and anything walking the result twice redoes the transform.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5811,6 +5811,34 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   </details>
 
+- <a id='prefer-lazy-map'></a>(<a href='#prefer-lazy-map'>link</a>) **Prefer `lazy.map` over `map` when the result is consumed in a single pass**.
+
+  <details>
+
+  <!-- ai-skill-include: generally autocorrectable, but still an important best practice -->
+
+  [![SwiftFormat: preferLazyMap](https://img.shields.io/badge/SwiftFormat-preferLazyMap-7B0051.svg)](https://swiftformat.info/rules/prerelease#preferLazyMap)
+
+  #### Why?
+
+  `map` allocates an array of every transformed element. When the only thing that happens to that array is a single walk over it, the allocation is pure waste: `lazy.map` transforms each element as it is visited instead. Operations that qualify are the ones that consume their receiver exactly once — `min`, `max`, `reduce`, `joined(separator:)` — plus `contains`, `allSatisfy`, and `first(where:)`, which can additionally stop early and skip transforming the rest.
+
+  ```swift
+  // WRONG
+  let minY = vertices.map { $0.y }.min()
+  let names = users.map { $0.name }.joined(separator: ", ")
+  let hasEmpty = rows.map { $0.title }.contains(where: { $0.isEmpty })
+
+  // RIGHT
+  let minY = vertices.lazy.map { $0.y }.min()
+  let names = users.lazy.map { $0.name }.joined(separator: ", ")
+  let hasEmpty = rows.lazy.map { $0.title }.contains(where: { $0.isEmpty })
+  ```
+
+  Reach for `lazy` only when the result really is consumed once. `sorted()` has to materialize its receiver anyway, and anything that walks the result more than once will redo the transform on each pass.
+
+  </details>
+
 **[⬆ back to top](#table-of-contents)**
 
 ## Apple Frameworks
--- a/README.md
+++ b/README.md
@@ -5821,19 +5821,19 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   #### Why?
 
-  `map` allocates an array of every transformed element. When the only thing that happens to that array is a single walk over it, the allocation is pure waste: `lazy.map` transforms each element as it is visited instead. Operations that qualify are the ones that consume their receiver exactly once — `min`, `max`, `reduce`, `joined(separator:)` — plus `contains`, `allSatisfy`, and `first(where:)`, which can additionally stop early and skip transforming the rest.
+  `map` allocates an array of every transformed element. When the only thing that happens to that array is a single walk over it, the allocation is pure waste: `lazy.map` transforms each element as it is visited instead. With an operation that can exit early, it also stops transforming as soon as it has an answer — the eager version below transforms every row before looking at any of them, even if the first row is the empty one it is searching for.
 
   ```swift
   // WRONG
-  let minY = vertices.map { $0.y }.min()
-  let names = users.map { $0.name }.joined(separator: ", ")
   let hasEmpty = rows.map { $0.title }.contains(where: { $0.isEmpty })
+  let minY = vertices.map { $0.y }.min()
 
   // RIGHT
-  let minY = vertices.lazy.map { $0.y }.min()
-  let names = users.lazy.map { $0.name }.joined(separator: ", ")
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -5811,6 +5811,34 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   </details>
 
+- <a id='prefer-lazy-map'></a>(<a href='#prefer-lazy-map'>link</a>) **Prefer `lazy.map` over `map` when the result is consumed in a single pass**.
```
calda: This first sentence is the only context AI will see, so maybe we also include some examples here: 
```suggestion
- <a id='prefer-lazy-map'></a>(<a href='#prefer-lazy-map'>link</a>) **Prefer `lazy.map` over `map` when the result is consumed in a single pass (`contains`, `min`, `max`, `joined`, etc).
```

jparise: Done — good catch on the AI context, I hadn't clocked that the headline is all that gets extracted.

Two small deviations from the literal suggestion:

- Closed the `**` (the suggestion left the bold unterminated) and kept the period outside it, matching the acronyms entry's `(`ID`, `URL`, etc)` style.
- Wrote `joined(separator:)` rather than `joined`, and added `reduce`. The separator matters precisely *because* the headline is the only context: bare `joined()` is deliberately not rewritten, since on a lazy sequence it resolves to the sequence-flattening overload and silently turns a `String` into a lazy sequence (that was one of the fixes in nicklockwood/SwiftFormat#2661 — it corrupted a macro's generated source rather than failing to compile). A headline reading just `joined` seemed likely to be taken as license to rewrite it.

calda: To me it's slightly difficult to understand what "when the result is consumed in a single pass" means. To brainstorm, maybe something like "when the chain of operations produces a single result (e.g. `contains`, `min`, `max`, `joined`, etc)". What do you think?

I think we want to capture somehow that `map + contains` (included) is different from `map + filter` (not included). "Consumed" maybe gets at that but is on the technical side

README.md
```diff
@@ -5811,6 +5811,34 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+
+  `map` allocates an array of every transformed element. When the only thing that happens to that array is a single walk over it, the allocation is pure waste: `lazy.map` transforms each element as it is visited instead. With an operation that can exit early, it also stops transforming as soon as it has an answer — the eager version below transforms every row before looking at any of them, even if the first row is the empty one it is searching for.
+
+  ```swift
+  // WRONG
+  let hasEmpty = rows.map { $0.title }.contains(where: { $0.isEmpty })
+  let minY = vertices.map { $0.y }.min()
+
+  // RIGHT
+  let hasEmpty = rows.lazy.map { $0.title }.contains(where: { $0.isEmpty })
```
calda: Hmmm, I suppose this specific example would be better expressed as:

```suggestion
  let hasEmpty = rows.contains(where: { $0.title.sEmpty })
```

We probably want an example that can't be simplified in that way. Or maybe this means `contains` isn't actually the best example to lead with. wdyt?

jparise: You're right, and I think it generalizes: `xs.map(f).contains(where: p)` always collapses to `xs.contains { p(f($0)) }`, and the same holds for `allSatisfy` and `first(where:)`. So *every* short-circuiting example is one where `.lazy` isn't the best available fix — which makes them all weak exemplars, not just this one.

The two cases with no better form are:

- `joined(separator:)` — needs an actual sequence of strings, nothing to fold into.
- `min`/`max` — the alternative is `vertices.min { $0.y < $1.y }?.y`, which is exactly the shape you pushed back on in nicklockwood/SwiftFormat#2655.

So I've led with those two and moved `contains` into a second example that makes the opposite point: fold the transform into the predicate, and there's no `map` left to make lazy. The short-circuiting operations stay in the list, since the rule does still improve those chains if you write them — but the guide shouldn't teach them as the motivating case.

README.md
jparise: Agreed, "consumed in a single pass" was doing too much work. Went with your framing, lightly adjusted:

> **Prefer `lazy.map` over `map` when the chain reduces to a single result (`joined(separator:)`, `min`, `max`, `reduce`, `contains`, etc)**

"Reduces to a single result" over "produces a single result" only because it points at *why* `filter` is excluded — `filter` does produce a single result, it's just a sequence rather than a value. The body now says that explicitly: it doesn't apply to operations producing another sequence, because the transformed elements get used more than once (or, for `sorted()`, have to be materialized regardless).

Reordered the list to put `joined(separator:)` and `min` first, matching the examples, for the reasons in the other thread.

README.md
calda: Perfect thanks!

README.md
```diff
@@ -5811,6 +5811,42 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+  // WRONG
+  let names = users.map { $0.name }.joined(separator: ", ")
+  let minY = vertices.map { $0.y }.min()
+
+  // RIGHT
+  let names = users.lazy.map { $0.name }.joined(separator: ", ")
+  let minY = vertices.lazy.map { $0.y }.min()
+  ```
+
+  This applies to any operation that turns the sequence into a single value: `joined(separator:)`, `min`, `max`, `reduce`, and also `contains`, `allSatisfy`, and `first(where:)`, which stop as soon as they have an answer. It does not apply to an operation that produces another sequence, like `filter` or `sorted()`, since the transformed elements are needed more than once — or, in `sorted()`'s case, have to be materialized anyway.
```
calda, line 5836:
```suggestion
  This applies to any operation that turns the sequence into a single value: `joined(separator:)`, `min`, `max`, `reduce`, and also `contains`, `allSatisfy`, and `first(where:)`, which stop as soon as they have an answer. It does not apply to an operation that produces another sequence, like `filter` or `sorted()`.
```

I don't think the " since the transformed elements are needed more than once" reason is quite right, the reason in my mind is because `lazy` changes the resulting collection type in those cases. I think it's ok to not go into much detail about that part though.

README.md
```diff
@@ -5811,6 +5811,42 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+
+  When the operation takes a predicate, folding the transform into that predicate is better still, because then there is no `map` to make lazy:
```
calda, line 5838: Good idea to mention this 👍🏻

calda (approved):
Love it thanks!

calda (approved):
Thanks!

PR #403, opened 2026-08-20
Refine testing expectation message guidelines

StarLard:
Add additional information about test expectation messages in order to improve the context captured in test failures.

#### Summary

Adds additional details to explain not just when you _shouldn't_ use expectation messages, but also when you _should_ use them and examples of how to use them effectively. When investigating a test failure, especially in an unfamiliar area of the codebase,  expectations such as `#expect(user == nil)` are often very difficult to troubleshoot. The expectation establishes what is expected, but it tells the debugger nothing about _why_ that expectation is there or _when_ it is expected to hold true. When triaging a test failure, `#expect(user == nil)` also fails to communicate the severity of the failure. In contrast, an effective expectation message, such as "Expected user not to have been populated after calling logout", signals to the triaging engineer "uh oh, the user is leaking after logout, which could be a severe issue". Expectation messages are powerful tools to give clues that aid developers in attempting to establish why a test has failed and why that failure is or is not significant.

#### Reasoning

The existing guidelines for test expectation methods could be interpreted as advocating for generally omitting them rather than advocating for using them prescriptively, which I believe was the original intent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5318,8 +5318,8 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   <details>
 
-  A message that restates the condition adds nothing over what `#expect` already reports on failure. Explain *why* the condition was expected to hold — here, because the test just engaged the warp drive:
-  
+  A message that restates the condition adds nothing over what `#expect` already reports on failure. Explain _why_ the condition was expected to hold — here, because the test just engaged the warp drive:
+
   ```swift
   // WRONG: Restates the condition
   #expect(spaceship.isWarpDriveActive, "Warp drive should be active")
@@ -5328,7 +5328,7 @@ _You can enable the following settings in Xcode by running [this script](https:/
   ```
 
   A message can also capture the requirement behind the assertion, which the condition alone doesn't convey:
-  
+
   ```swift
   // WRONG: Restates the condition
   #expect(spaceship.speed > lightSpeed, "Speed should be greater than light speed")
@@ -5348,7 +5348,7 @@ _You can enable the following settings in Xcode by running [this script](https:/
   ```
 
   It is fine to omit expectation message strings when there is no additional context to add that isn't already clear from surrounding context. For example, it's not necessary to restate the name of the test case itself in an assertion message:
-  
+
   ```swift
   // WRONG: The message just repeats the test name
   @Test
--- a/README.md
+++ b/README.md
@@ -5314,25 +5314,54 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   </details>
 
-- <a id='avoid-redundant-expectation-comments'></a>(<a href='#avoid-redundant-expectation-comments'>link</a>) **In Swift Testing, avoid expectation message strings that restate the expectation without adding additional context.** Unlike `XCTAssert`, the Swift Testing `#expect` macro generates detailed failure messages that include the expectation condition.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5314,25 +5314,54 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   </details>
 
-- <a id='avoid-redundant-expectation-comments'></a>(<a href='#avoid-redundant-expectation-comments'>link</a>) **In Swift Testing, avoid expectation message strings that restate the expectation without adding additional context.** Unlike `XCTAssert`, the Swift Testing `#expect` macro generates detailed failure messages that include the expectation condition.
+- <a id='avoid-redundant-expectation-comments'></a>(<a href='#avoid-redundant-expectation-comments'>link</a>) **In Swift Testing, avoid expectation message strings that restate the expectation without adding additional context.** Unlike `XCTAssert`, the Swift Testing `#expect` macro generates detailed failure messages that include the expectation condition. When a message does add context, phrase it as the intent behind the assertion — the reason the condition was expected to hold — rather than a restatement. Avoid hardcoding expected values into the message so that the message can't drift from the assertion.
```
calda: Would be nice to avoid the `—`s here. Here is one potential wording:


```suggestion
- <a id='avoid-redundant-expectation-comments'></a>(<a href='#avoid-redundant-expectation-comments'>link</a>) **In Swift Testing, avoid expectation message strings that restate the expectation without adding additional context.** Unlike `XCTAssert`, the Swift Testing `#expect` macro generates detailed failure messages that include the expectation condition. When an expectation message string does add valuable context, phrase it as the intent behind the assertion (the reason the condition was expected to hold) rather than restating the condition. Avoid hardcoding expected values into the message so that the message can't drift from the assertion.
```

calda (commented):
Thanks! Let's get @jqsilver's thoughts as well.

No need to run a rule proposal for this since it's just tweaking the wording of the existing rule without fundamentally changing it.

README.md
```diff
@@ -5314,25 +5314,54 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+
+  ```swift
+  // WRONG: Hardcodes the expected value
+  #expect(spaceship.speed == expectedCruisingSpeed, "Cruising speed should be warp 5")
+
+  // RIGHT: Describes the value's significance so the message can't drift
+  #expect(spaceship.speed == expectedCruisingSpeed, "Cruise control should keep the spaceship at cruising speed")
+  ```
+
+  When a single-assertion test is named as a sentence, a message that repeats that name is redundant:
```
calda: Better to broaden this a bit since we aren't requiring all expectations have a message string. I so like using this specific example though because it's pretty obvious that the message is redundant.

```suggestion
  It is fine to omit expectation message strings when there is no additional context to add that isn't already clear from surrounding context. For example, it's not necessary to restate the name of the test case itself in an assertion message:
```

calda (approved):
I like the additions to this rule, thanks!

README.md
```diff
@@ -5314,25 +5314,54 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
 
   <details>
 
+  A message that restates the condition adds nothing over what `#expect` already reports on failure. Explain _why_ the condition was expected to hold — here, because the test just engaged the warp drive:
+
+  ```swift
+  // WRONG: Restates the condition
+  #expect(spaceship.isWarpDriveActive, "Warp drive should be active")
+  // RIGHT: Explains why the drive should be active
+  #expect(spaceship.isWarpDriveActive, "The warp drive should be active after calling engageWarpDrive()")
```
calda, line 5327: I know this is a toy example but this is a case where I don't think the comment adds value. 

"The warp drive should be active after calling engageWarpDrive()" is trivially obvious from a test case that asserts `#expect(spaceship.isWarpDriveActive)` after calling `engageWarpDrive()`:

```swift
engageWarpDrive()
#expect(spaceship.isWarpDriveActive)
```

PR #404, opened 2026-09-01
Sort `@_spi` imports last

calda:
#### Summary

This PR proposes an update to the "sort imports" rule, where we sort `@_spi` imports last, below any `@testable` imports.

  ```swift
  // WRONG

  import Foundation
  @testable import OrbitService
  import SolarSystemFeature
  @_spi(Internal) import GalaxyUI
  import Testing

  // RIGHT

  import Foundation
  import SolarSystemFeature
  import Testing
  @testable import OrbitService
  @_spi(Internal) import GalaxyUI
  ```

#### Reasoning

Similarly to sorting `@testable` imports last, this keeps the same kinds of imports grouped together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4532,7 +4532,7 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
 ## File Organization
 
-- <a id='alphabetize-and-deduplicate-imports'></a>(<a href='#alphabetize-and-deduplicate-imports'>link</a>) **Alphabetize and deduplicate module imports within a file.** Place all imports at the top of the file below the header comments. Do not add additional line breaks between import statements. Add a single empty line before the first import and after the last import.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4532,7 +4532,7 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
 ## File Organization
 
-- <a id='alphabetize-and-deduplicate-imports'></a>(<a href='#alphabetize-and-deduplicate-imports'>link</a>) **Alphabetize and deduplicate module imports within a file.** Place all imports at the top of the file below the header comments. Do not add additional line breaks between import statements. Add a single empty line before the first import and after the last import.
```
calda, line 4535: > Do not add additional line breaks between import statements.

We are not currently enforcing this (SwiftFormat doesn't have an option for this at the moment). It seems reasonable to include linebreaks between the regular imports, `@testable` imports, and `@_spi` imports if you want (we don't require that either though).

PR #409, opened 2026-09-16
Add rule: prefer struct for Swift Testing suites

miguel-jimenez-0529:
#### Summary

Enables SwiftFormat's `preferStructSwiftTestingSuites` rule ([SwiftFormat#2680](https://github.com/nicklockwood/SwiftFormat/pull/2680)) and adds the corresponding style-guide entry to the **Testing** section, next to the existing `redundantSwiftTestingSuite` entry, since both are about how a suite type is declared.

The rule is disabled by default upstream. No `Package.swift` change is needed — the currently pinned SwiftFormat build (`2026-09-03-c`) already includes it, since that nightly was cut ~54 minutes after the rule merged.

#### Reasoning

Swift Testing creates a fresh instance of the suite for every test case, so each case is already isolated from the others. A `struct` states that directly: no shared mutable instance, no inheritance to reason about, and no `final` needed to rule out subclassing.

```swift
// WRONG
final class SpaceshipTests {
  @Test
  func `warp drive enables FTL travel`() { ... }
}

// RIGHT
struct SpaceshipTests {
  @Test
  func `warp drive enables FTL travel`() { ... }
}
```

`enum` suites of only `static` members are isolated too, but can't gain `init` setup or instance `@Test` methods later without a rewrite, so the rule converts those as well (`@Suite enum` → `@Suite struct`).

This also matches idiomatic usage: every suite type on Apple's [Organizing test functions with suite types](https://developer.apple.com/documentation/testing/organizingtests) page is a `struct` — there are no `class` or `enum` suites anywhere in it. The entry cites that page.

(Handy trick for checking this kind of thing: appending `.md` to any Apple docs URL returns the page as markdown.)

#### What the rule leaves alone

Verified against the pinned build. The rule only applies to files that `import Testing`, and only to types it recognizes as suites (an `@Suite` attribute, or a body containing an `@Test` method). Beyond that it skips:

- **Classes with a mutable stored instance `var`.** A `@Test` method isn't `mutating`, so a suite whose cases assign to stored properties can't be a `struct`. This is the one skip case the style-guide entry documents, since it's the one that reflects a real constraint rather than rule conservatism.
- **Classes with any conformance or superclass.** Conservative — a `struct` can conform to protocols fine, but the rule can't tell a protocol from a base class, so it declines both.
- **Enums with `case` declarations**, which aren't suites in the first place.

Computed `var`s, `static var`s, and `let`-only stored properties don't block conversion, which is correct — none of them require reference semantics.

#### Caveat worth flagging: `deinit`

A suite class whose only class-requiring member is a `deinit` **is** converted, producing code that doesn't compile:

```swift
// Input
final class DeinitTests {
  @Test func `resource is released`() { ... }
  deinit { Observatory.shared.reset() }
}

// Output — error: deinitializer cannot be declared in struct 'DeinitTests' that conforms to 'Copyable'
struct DeinitTests {
  @Test func `resource is released`() { ... }
  deinit { Observatory.shared.reset() }
}
```

`deinit` isn't part of the rule's skip conditions upstream. **This is not theoretical — it has already been hit in `airbnb/apps`**, on a suite that used `deinit` to unmock a singleton, and the file had to be exempted to get the migration through.

So this needs an upstream fix before we roll the rule out broadly. Happy to file that against SwiftFormat and hold this PR until it lands, or land the style-guide entry now and enable the rule in a follow-up — whichever reviewers prefer.

Note that the style-guide entry deliberately does **not** show a `deinit` suite as an approved pattern; its `class` example uses mutable stored state instead, which is the constraint that genuinely requires reference semantics.

#### Verification

- `swift package format --lint` — clean (this repo's own tests use XCTest with no `import Testing`, so the rule doesn't touch them)
- `swift test` — 8 tests pass
- `npx prettier --check README.md` — clean
- Site generation checked locally: the entry renders on the site and is correctly excluded from the AI skill, since it's autocorrectable and carries a SwiftFormat badge

#### Not yet done

Not yet applied across the Airbnb iOS codebase — happy to run the migration and link the `airbnb/apps` PR here, as with the recent rule additions.


README.md
```diff
@@ -5304,6 +5304,66 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+
+  <details>
+
+  [![SwiftFormat: preferStructSwiftTestingSuites](https://img.shields.io/badge/SwiftFormat-preferStructSwiftTestingSuites-7B0051.svg)](https://swiftformat.info/rules/prerelease#preferStructSwiftTestingSuites)
+
+  #### Why?
+
+  Swift Testing creates a fresh instance of the suite for every test case, so each case is already isolated from the others. A `struct` states that directly: there is no shared mutable instance, no inheritance to reason about, and no need for `final` to rule out subclassing.
+
+  An `enum` of `static` members is isolated too, but it cannot gain `init` setup or instance `@Test` methods later without being rewritten. A `struct` starts where an `enum` would end up.
```
calda: I think we can skip the discussion about `enum`s, and drop the `enum` WRONG example, to simplify the content a bit. The phrasing also has a few Claude-isms.

README.md
```diff
@@ -5304,6 +5304,66 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
 
+- <a id='prefer-struct-swift-testing-suites'></a>(<a href='#prefer-struct-swift-testing-suites'>link</a>) **Prefer `struct` for Swift Testing suites.** Swift Testing instantiates the suite type once per test case, so suites rarely need reference semantics. Reach for a `class` only when the suite actually requires one.
+
+  <details>
+
+  [![SwiftFormat: preferStructSwiftTestingSuites](https://img.shields.io/badge/SwiftFormat-preferStructSwiftTestingSuites-7B0051.svg)](https://swiftformat.info/rules/prerelease#preferStructSwiftTestingSuites)
+
+  #### Why?
+
+  Swift Testing creates a fresh instance of the suite for every test case, so each case is already isolated from the others. A `struct` states that directly: there is no shared mutable instance, no inheritance to reason about, and no need for `final` to rule out subclassing.
```
calda: Also quite a few Claude-isms here. I think the content in the bullet is probably sufficiently clear that we don't need a `Why?` here

README.md
```diff
@@ -5304,6 +5304,66 @@ _You can enable the following settings in Xcode by running [this script](https:/
 
   </details>
 
+- <a id='prefer-struct-swift-testing-suites'></a>(<a href='#prefer-struct-swift-testing-suites'>link</a>) **Prefer `struct` for Swift Testing suites.** Swift Testing instantiates the suite type once per test case, so suites rarely need reference semantics. Reach for a `class` only when the suite actually requires one.
```
```suggestion
- <a id='prefer-struct-swift-testing-suites'></a>(<a href='#prefer-struct-swift-testing-suites'>link</a>) **Prefer defining Swift Testing suites as `struct`s.** Swift Testing instantiates the suite type once per test case, so suites rarely need reference semantics.
```

calda: Trying to make the wording a bit more clear and reduce Claude-isms ^

calda (approved):
Thanks! 😃

README.md
```diff
@@ -5304,6 +5304,68 @@ _You can enable the following settings in Xcode by running [this script](https:/
...
+  struct TelescopeTests {
+    @Test
+    static func `telescope tracks target`() { ... }
+  }
+  ```
+
+  A `@Test` method is not `mutating`, so a suite whose test cases assign to stored properties has to stay a `class`:
+
+  ```swift
+  // ALSO RIGHT: the test case assigns to stored state, which a non-mutating `@Test` method can't do in a struct.
```
calda: I think this is a Swift Testing anti-pattern, so I think it's better we don't include it at all.

Have been thinking about how to include the `<details>` content in the Swift Testing skills. `ALSO RIGHT` code with anti-patterns is more confusing to AI.

Of course this specific rule probably doesn't need to be in the AI skill since it's pretty auto-correctable.

calda: Alternatively, we could include it with a caveat like "However, mutable test suite instance properties are usually an anti-pattern in Swift Testing".

